PR #13, opened 2016-01-28
Coding standards

BronsonQuick:
Hey @modemlooper, I had a few people ask me about why you use Dockunit instead of Travis so I've PRed to add Travis instead and fixed up some things it picked up.

I'm still getting my head around the functions in the plugin generator you use but I'll get more familiar with it soon enough :+1:


ntwb:
What about `bower.json`, presuming that could probably also go unless someone has plans to use it?

BronsonQuick:
@ntwb I actually meant to remove that and forgot! Thank :dancers:

modemlooper:
Sounds good, should use travis as thats what bp core uses.

PR #17, opened 2016-01-29
DO NOT MERGE - LOOK BUT DON'T TOUCH

ntwb:
I just removed the `node_modules` from my repo fork Git history this is the resultant pull request! :gift: 

@boonebgorges tip was to use `filter-branch` via Slack [here](https://wordpress.slack.com/archives/buddypress/p1453941417000395)

With said tip I followed the answer here: http://stackoverflow.com/a/17824718

And ran (We'd already added `node_modules` to `.gitignore`):
```
git filter-branch --tree-filter 'rm -rf node_modules' --prune-empty HEAD
git gc
git push origin master --force
```

The result is this pull request, everything looks good to me, but do not rely solely on my opinion here, I may have missed something :smirk: 

Maybe rather doing this via a pull request it should be done directly on the main repo?

As Boone noted:  
> "... you’ll break existing hashes. If you’re going to do that, do it now before anyone is using the repo"

What do you think @modemlooper @BronsonQuick @boonebgorges ?


modemlooper:
you should be able to merge master into your fork and it should remove it

ntwb:
> you should be able to merge master into your fork and it should remove it
>
> https://help.github.com/articles/syncing-a-fork/

Not sure what you mean here, my fork is (or was) up to date with `buddypress/BP-REST` when I created this PR, it also includes an upstream remote to `buddypress/BP-REST`.

I created this pull request so that the ~1,000 `node_modules` objects that remain in the Git repos history can be removed. This repo is incredibly slow to clone because of this history in the repo, so removing it from the history would fix that.

My suggestion here is to if everyone is in agreement then someone with commit here would then run the same commands I included in the original PR message directly on the `master` branch and commit them directly to `master` and push those changes up here to GitHub. From my quick reading of `filter-branch` and after a `-- force` push merging a pull request like this one is troublesome at best, hence a committer should do this directly and not try to merge this pull request.

boonebgorges:
Given that the repo only has 4 forks - all of which belong to friends of the project - this history rewrite should be done now, if it's going to be done at all.

@ntwb Thanks for experimenting with this. However, your branch doesn't actually remove all references from history, meaning that the pack is still quite large. Here are some more thorough instructions that I used locally to remove `node_modules` completely from the repo history: http://www.link-intersystems.com/blog/2014/07/17/remove-directories-and-files-permanently-from-git/ This process gets the repo size from 2.1M to 316K.

@BronsonQuick I personally lean toward doing this - and pinging all forks to rebase/reclone - but I'll leave the final decision up to you. Having `node_modules` in the repo history is annoying, but not the end of the world. If you'd rather avoid the trouble, it's OK with me.

ntwb:
Damn and drat :smirk: 

Thanks for that @boonebgorges, just ran through it step by step, and then the all-in-one :+1: 

Following the guide linked, cut and paste the follwing snippets, first export an environment variable:
```bash
export GIT_DELETE_PATH_BP_REST="node_modules"
```

Remove all the `node_modules` things:
```sh
git filter-branch \
    --index-filter "git rm -rf --cached --ignore-unmatch $GIT_DELETE_PATH_BP_REST" \
    --tag-name-filter "cat" \
    -- \
    `git log --reverse -- $GIT_DELETE_PATH_BP_REST | head -n 1 | awk '{ print $2 "^1" }'`..`git log -1 -- $GIT_DELETE_PATH_BP_REST | head -n 1 | awk '{ print $2 }'` \
    --all \
&& \
git for-each-ref --format="%(refname)" refs/original/ | xargs -n 1 git update-ref -d \
&& \
git reflog expire --expire=now --all \
&& \
git gc --prune=now
```

Which will result in an output of ~6,000 lines of references to deleting files and folders from various `node_modules` paths and look similar to https://gist.github.com/ntwb/b9e81eebac997f41cb5e

Force push those changes back here to GitHub:
```sh
git push origin *:* --force
```

Lastly ping the forks to rebase or re-clone :fork_and_knife:

paulgibbs:
I've gone ahead and done this.

PR #19, opened 2016-04-26
Remove support for PHPUnit.

paulgibbs:
We don't intend to unit test the REST API endpoints, so this is unnecessary.


danielbachhuber:
> We don't intend to unit test the REST API endpoints

Why not?

paulgibbs:
We have (or will have) integration tests for the BuddyPress API functions that the REST API implementation will use.

My preference will be to implement some kind of functional testing against a test site. Perhaps something custom using Mink. Or maybe something similar to Behat (I am not sure but I am not convinced Behat would be the right tool for this, though it'd be capable). But hopefully there's some sort of tooling we can adopt from other projects to do this.

danielbachhuber:
Have you seen the [style of tests](https://github.com/WP-API/WP-API/tree/develop/tests) we've used for the WP REST API plugin? Aside from https://github.com/WP-API/WP-API/issues/1757, it's served us quite well. I think there's an advantage to using something PHPunit-based, so other WP devs have an easier starting point.

PR #22, opened 2016-04-27
Plugin loader improvements

paulgibbs:
Removed a lot of the generator-plugin-wp bootstrap to simplify loading.
Also hooked loading to new `bp_rest_api_init` action which I sneaked into core earlier today.

Once this stuff gets into BuddyPress we'll end up using an autoloader to load these classes as/when required so I kept the `require` because it was already there and it's fine for now,


BronsonQuick:
🎉

PR #30, opened 2016-07-12
Group READABLE endpoint

dcavins:
First pass at allowing fetching of groups using REST. Fetches single
groups and collections of groups using an argument format much like
`groups_get_groups()`.

Still need to work out accepting meta arguments and also visibility of hidden groups.

How are we going to test all of this REST stuff?

Finally, is the REST `context` parameter really a security thing? It appears to be used that way in the main plugin, but I'm doubtful. I've used it here to make `populate_extras` only be possible with the context elevation, but it's still really a view context with increased user permissions, so maybe that's incorrect.

### DIFF
--- a/bp-rest.php
+++ b/bp-rest.php
@@ -48,5 +48,11 @@ function bp_rest_api_endpoints() {
 		$controller = new BP_REST_Activity_Controller();
 		$controller->register_routes();
 	}
+
+	if ( bp_is_active( 'groups' ) ) {
+		require_once( dirname( __FILE__ ) . '/includes/bp-groups/classes/class-bp-groups-endpoints.php' );
+		$controller = new BP_REST_Groups_Controller();
+		$controller->register_routes();
+	}
 }
 add_action( 'bp_rest_api_init', 'bp_rest_api_endpoints' );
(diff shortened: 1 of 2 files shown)


BronsonQuick:
Awesome! Thanks for making a start on this!

>How are we going to test all of this REST stuff?

I originally had Travis for this but @paulgibbs chose to remove it: 
https://github.com/buddypress/BP-REST/commit/0ee3de4df79005dc48d55b01df36c331f8180c20

>Finally, is the REST `context` parameter really a security thing?

I believe it's to do with roles and permissions.

paulgibbs:
Don't know about testing yet. Going to talk to Boone and John this Monday and we'll come up with a proposal.

`context` is used as a way of managing fields returned with objects on certain endpoints, for any reason. It's not meant to manage any kind of privileged data access. Generally, it seems pretty situation specific, and I don't think we need to use it until we find somewhere it'd help.

cezarneaga:
`wp-json/buddypress/v1/groups` doesn't work for me. does it work for you?

renatonascalves:
@cezarneaga Could you elaborate on how it doesn't work for you? Do you see any errors?

cezarneaga:
sorry. no errors just an empty array. groups/[id] works fine though

renatonascalves:
Thank you @dcavins 

I reviewed your PR and fixed what's necessary. This is good to merge!
I'll add the unit tests later. :)

PR #54, opened 2017-12-13
Filter the link returned in the WP API users response.

dcavins:
Replace the standard WP "author" link with a link to the BuddyPress user domain.


renatonascalves:
Thank you @dcavins 💃

PR #67, opened 2018-06-11
Add notification endpoint v1

baconbro:
Add notification support for Bp rest API. The use of function bp_notifications_get_notifications_for_user() rather than method get($args) from notification classes is due to the plus value of having a content return rather than pure notification results.


README.md
```diff
@@ -28,6 +28,7 @@ Please use and provide feedback!
 - [ ] Emails
 - [ ] Friends
 - [ ] Signups
+- [x] Notifications
```
renatonascalves: Could you add "Partial Support"? So that people know that this endpoint doesn't have full support yet.

includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
```diff
@@ -0,0 +1,225 @@
...
+	/**
+	 * Retrieve notifications.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Request Notification object data.
+	 */
+	public function get_items( $request ) {
+		$notifications = bp_notifications_get_notifications_for_user( bp_loggedin_user_id(), $format = 'object' );
```
renatonascalves: What if I want to get notifications from other users? This endpoint must support the all the options the `get()` method support. 

https://github.com/buddypress/BuddyPress/blob/d1cb62a361d029d4fcf1c3d5c62b885edd63c6a5/src/bp-notifications/classes/class-bp-notifications-notification.php#L636

This will allow more flexibility for people who want to get items based on some criteria.

I noticed your first implementation had this support but you then removed it.

includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
```diff
@@ -0,0 +1,225 @@
...
+	 * Check if a given request has access to notification items.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function get_items_permissions_check( $request ) {
+		// Bail early.
+		if ( ! $this->can_see() ) {
```
renatonascalves: The notification id missing here.

includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
baconbro: Yes, it's because of the content. With get(), you only get pur data, without  content. So you can't display the notification message.
The other consideration is about privacy, a user should only see his/her notifiction.
I do agree that it might be better but I suggest in a future version.

includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
baconbro: Yes, same idea here, it was plan to be implanted but no need now because privacy is taken care by only present login user's notification.
Will be useful in next version.

README.md
baconbro: Yes for sure!

renatonascalves:
@baconbro I'm going to take a look at it. By the way, I noticed you are working on a friends component integration.

Can you send a PR with it when you are finished? That'd be useful as well. :)

renatonascalves:
I'll merge this and polish a little bit.

renatonascalves:
@baconbro Great! I saw now you added more code to your version. I'll take a look and "steal" the good parts. =)

PR #71, opened 2018-08-20
Fix Activity delete permission failing unittest

imath:
Hi @renatonascalves 

I will play around with the BP Rest API. So here's my first contribution :)

Unittest fix + fixes the fact `BP_REST_Activity_Endpoint->delete_item_permissions_check()` is too permissive as any logged in user could delete any activity item.


renatonascalves:
@imath Great! Thank youuu! 💃

PR #72, opened 2018-08-21
Improve the BP_REST_Activity_Endpoint->show_hidden() method

imath:
This method is used in the get_items() one to know if hide_sitewide activities should be fetched for the loggedin user.
1. As the default value of the primary id is an array, in order to avoid errors when checking if the user is a member/admin/mod of the group we need to extract the first value of this array.
2. Public Groups activities should be fetchable for any logged in user
3. A filter makes it possible for custom components to add their own logic about the hide_sitewide param.
4. If statements have been reorganized to avoid unnecessary checks.
5. bp_loggedin_user_id() does not seem to be set during a REST API request, get_current_user_id() is.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -120,11 +120,13 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			$args['filter']['primary_id'] = $item_id;
 
 			/**
+			 * Make sure the Item ID is an integer.
+			 *
 			 * @todo Check why the primary_id default's value is an array
 			 *       in $this->get_collection_params()?
 			 */
 			if ( is_array( $item_id ) ) {
-				$item_id = reset( $item_id );
+				$item_id = (int) reset( $item_id );
 			}
 		}
 
@@ -792,7 +794,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			if ( groups_is_user_admin( $user_id, $item_id ) || groups_is_user_mod( $user_id, $item_id ) ) {
 				$retval = true;
 
-			// User is a member of the group.
+				// User is a member of the group.
 			} elseif ( (bool) groups_is_user_member( $user_id, $item_id ) ) {
 				$retval = true;
 			}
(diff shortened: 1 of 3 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -114,8 +114,18 @@ public function get_items( $request ) {
 			$args['filter']['action'] = $request['type'];
 		}
 
+		$item_id = 0;
 		if ( isset( $request['primary_id'] ) ) {
-			$args['filter']['primary_id'] = $request['primary_id'];
+			$item_id                      = $request['primary_id'];
+			$args['filter']['primary_id'] = $item_id;
```
renatonascalves: If it is returning an array, this liine should gone after reset you are doing below. No?!

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -114,8 +114,18 @@ public function get_items( $request ) {
...
+
+			/**
+			 * @todo Check why the primary_id default's value is an array
+			 *       in $this->get_collection_params()?
+			 */
+			if ( is_array( $item_id ) ) {
```
renatonascalves: I'll do some tests to understand why this is happening.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
imath: Well actually I don’t touch to the $request[‘primary_id’], i’m just catching it into $item_id to get its first index for the show_hidden() function, meaning $args[‘filter’][‘primary_id’] is unchanged.

I ran the full Tests suite successfully  👌

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
imath: It’s an hardcoded setting -> default value in the collection params is set to array(), I’ve left the way it was, but i’m just wondering the reason as the item_id should be an integer imho.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
renatonascalves: @imath But if $item_id is an array, the value being added to primary_id, is an array instead of an int...

renatonascalves: The code on get_collection_params is wrong, that should return an array. I'll update that.

imath: Yes nothing changes it's still an array and `$args[‘filter’][‘primary_id’]` is still an array. But $item_id is reset after to and integer so that BP_REST_Activity_Endpoint()->show_hidden() receives an integer as it always should have 👍 

Thanks for the merge ;)

PR #75, opened 2018-08-22
The user_login is a required field to insert a new user.

imath:
The BP_REST_Members_Endpoint class extends the WP_REST_Users_Controller one but has a different item schema. In particular, it uses an unrequired user_login key where the parent is using a required username key. As a result creating an item is generating a PHP error in the wp_insert_user() function at line 1526.
The fix is making the user_login key required, and extends the prepare_item_for_database() method so that the user_login is successfully prepared.


renatonascalves:
@imath Great! Thank you for fixing this. :)

PR #76, opened 2018-08-22
Make BP_REST_Activity_Endpoint->create_item() more friendly to BP plugins

imath:
Using the method as a plugin is quite complex as it requires to filter bp_activity_add() making sure a rest request is being performed. I suggest to be nicer with plugins wishing to use the BP Rest API and in particular this activity endpoint by:
Allowing custom components to create items, for instance I see no reason to restrict Activity creation to Core components, it should be available for all active components.
It should allow hidden activities to be created, the Group will need it anywayfor the ! public statuses.

This PR also improves some parts like:
- use a default value for the user ID as bp_loggedin_user_id() is not set in activity functions when creating Group Activities (get_current_user_id() is now used as the default value)
- add the missing unit tests for the create item method when posting group activities.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -266,14 +266,24 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 		$id                = $request['id'];
 		$parent            = $request['parent'];
 		$type              = $request['type'];
+		$show_hidden       = $request['hidden'];
 		$activity_id       = 0;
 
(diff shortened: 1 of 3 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -1069,7 +1117,7 @@ public function get_collection_params() {
 		$params['component'] = array(
 			'description'       => __( 'Limit result set to items with a specific BuddyPress component.', 'buddypress' ),
```
renatonascalves: I think this description should be updated to make it clear it is active components now.

renatonascalves:
I think the suggestions are really good. 🗡

PR #78, opened 2018-08-24
Add Response headers about total number of activities and pages

imath:
Client scripts need to be informed about the total amount of activities found for the request and about the total amount of available pages for the results of the request. Otherwise building a pagination is not possible and it needs the script to keep on requesting untill the response returns no results.
To avoid altering the response data with these two piece of informations, I suggest to use the allowed headers made available by the WordPress REST API :
- X-WP-Total is informing about the total amount of activities found.
- X-WP-TotalPages is informing about the number of pages.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -79,7 +79,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return WP_REST_Request List of activities object data.
+	 * @return WP_REST_Response List of activities response data.
 	 */
 	public function get_items( $request ) {
 		$args = array(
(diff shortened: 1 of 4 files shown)


imath:
@renatonascalves if you're fine with this suggestion I think the other endpoints' `get_items()` methods should probably use the same way of informing about paginated results.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -141,20 +141,29 @@ public function get_items( $request ) {
 			);
 		}
 
-		$retval = rest_ensure_response( $retval );
+		$response = rest_ensure_response( $retval );
+
+		// Set headers to let the Client Script be aware of the pagination.
+		if ( ! empty( $activities['total'] ) && $args['per_page'] ) {
```
renatonascalves: @imath Is there a way we can create a method that does that? We would only add the attributes. That would avoid duplicating this code in the other components.

My concern is if the other components have different way to paginate...

imath:
Hi @renatonascalves 

Thanks a lot for your review. Very good point about code duplication. I've updated the PR so that it now uses a function that can be used by any BP Endpoint.

As far as I know all components are paginating the results the same way.

renatonascalves:
@imath Great! That's a great addition to the plugin.

PR #79, opened 2018-08-28
Activities need to be rendered

imath:
Instead of having the raw data fetched into the view context, I think we should provide rendered activities into this context and provide both (raw data + rendered activity) into the edit context.
In other words, we need to make sure the activity content filters are fired to sanitize the output/run formatting filters when the context is a view one.
I suggest to do just like WordPress does for the Post Types : using an object type for the content key of the schema with raw and rendered properties.
From the client side, developers will need to update their scripts to display the content and use data.content.rendered instead of data.content for instance.

FYI I've added a test about embeds into activity content, to avoid a fatal error now WordPress is using a build/src architecture I've edited the phpunit.xml.dist file to make sure the `SCRIPT_DEBUG` constant is false. As we are not testing JavaScript, I don't think it's a problem.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -559,20 +559,68 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	}
 
 	/**
+	 * Renders the content of an activity.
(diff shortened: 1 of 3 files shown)


renatonascalves:
@imath Great addition as it aligns the output to how the bp-graphql is doing right now.

🍡

PR #80, opened 2018-08-30
Add Activity favorites

imath:
The idea is to implement it a bit differently than how it is done in AJAX. The user is updating their favorites for a specific activity Rest request.
When fetching activities so that it is easy for a client to know whether an activity is favorited or not, i have updated the schema with a readonly property (favorited).
I am only getting the user favorites for get_items/get_item and update_favorite methods as i dont think we need it elsewhere

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -21,8 +21,9 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @since 0.1.0
 	 */
 	public function __construct() {
-		$this->namespace = 'buddypress/v1';
-		$this->rest_base = buddypress()->activity->id;
+		$this->namespace      = 'buddypress/v1';
+		$this->rest_base      = buddypress()->activity->id;
+		$this->user_favorites = array();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -201,6 +214,25 @@ public function get_item( $request ) {
...
+	 * @return boolean True if the user is logged in. False otherwise.
+	 */
+	public function get_user_permission_check() {
+		if ( ! is_user_logged_in() ) {
+			return false;
+		}
+
+		// Current user favorites.
+		$user_favorites       = bp_activity_get_user_favorites( get_current_user_id() );
+		$this->user_favorites = array_filter( wp_parse_id_list( $user_favorites ) );
```
renatonascalves: Maybe change here for `$user_favorites['activities']`. And confirm if it is not empty.

I'm not sure if `array_filter` returns `false` on an empty array.

renatonascalves: Confirm if it is not empty. I'm not sure if `array_filter` returns `false` on an empty array.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -557,6 +598,108 @@ public function delete_item_permissions_check( $request ) {
...
+				__( 'Sorry, you need to be logged in to update your favorites.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		}
+
+		if ( ! bp_activity_can_favorite() ) {
+			return new WP_Error( 'rest_activity_cannot_favorite',
+				__( 'Sorry, Activities cannot be favorited.', 'buddypress' ),
```
renatonascalves: Activity, not Activities. We are not bulk-updating activities, so singular.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -629,6 +772,7 @@ public function prepare_item_for_response( $activity, $request ) {
 			'status'                => $activity->is_spam ? 'spam' : 'published',
 			'title'                 => $activity->action,
 			'type'                  => $activity->type,
+			'favorited'             => in_array( $activity->id, $this->user_favorites, true ),
```
renatonascalves: That's interesting!

renatonascalves:
Thanks @imath :)

Issue #81, opened 2018-08-31
Fix Hidden Groups Permissions Check


tannerm:
User's should be able to get hidden groups that they are members of but the `bp_moderate` check is overwriting that.

PR #82, opened 2018-08-31
Update hidden groups permissions check

tannerm:
Allow users who can't bp_moderate to get hidden groups that they are a member of. #81

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -339,12 +339,8 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 		$retval = true;
 
 		// Only bp_moderators and logged in users (viewing their own groups) can see hidden groups.
-		if ( ! empty( $request['show_hidden'] ) ) {
+		if ( ! empty( $request['show_hidden'] ) && ! bp_current_user_can( 'bp_moderate' ) ) {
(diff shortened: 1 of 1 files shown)


includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -339,14 +339,10 @@ protected function can_see( $request, $edit = false ) {
 		$retval = true;
 
 		// Only bp_moderators and logged in users (viewing their own groups) can see hidden groups.
-		if ( ! empty( $request['show_hidden'] ) ) {
+		if ( ! empty( $request['show_hidden'] ) && ! bp_current_user_can( 'bp_moderate' ) ) {
```
renatonascalves: @tannerm Indeed it is a bug. But it's better to change from negative to positive.

Right now, you are not allowing moderators to see it. Something like that will fix it allowing access for both.

```
$retval = false;

// Only bp_moderators and logged in users (viewing their own groups) can see hidden groups.
if ( ! empty( $request['show_hidden'] ) ) {
	if ( ! empty( $request['user_id'] ) && bp_loggedin_user_id() === $request['user_id'] ) {
		$retval = true;
	}

	if ( bp_current_user_can( 'bp_moderate' ) ) {
		$retval = true;
	}
}
```

Can you update it?

tannerm: @renatonascalves thanks for resolving this! Sorry it took so long to reply.

renatonascalves: No problem! :)

PR #91, opened 2018-10-03
Notifications: Update Support

renatonascalves:
This PR updates the notification endpoint to update a notification marking it as read or unread.

### DIFF
--- a/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
+++ b/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
@@ -59,6 +59,12 @@ class BP_REST_Notifications_Endpoint extends WP_REST_Controller {
 				),
 			),
 			array(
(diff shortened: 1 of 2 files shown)


includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
```diff
@@ -290,6 +296,103 @@ public function create_item_permissions_check( $request ) {
...
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function update_item( $request ) {
+		$notification = $this->get_notification_object( $request );
+
+		$updated = BP_Notifications_Notification::update(
```
renatonascalves: This is not updating the status of the notification. :/

PR #109, opened 2018-12-07
Don't use old function signature for `groups_get_group()`.

boonebgorges:
It was changed in https://buddypress.trac.wordpress.org/changeset/11090.

@renatonascalves Submitted for your perusal :-D


boonebgorges:
Thank you!

PR #111, opened 2018-12-08
Groups: verify if user has access to hidden groups

### LINKED ISSUE #107: Site admins cannot query show_hidden=false
The following line: https://github.com/buddypress/BP-REST/blame/551da0dd50cbde89a2816a8551409b53534548d6/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php#L105

makes it impossible for site admins to query for groups with the `show_hidden=false` param. This doesn't feel right - I can think of many instances where you might want to. I think it's fine to have the *default* value of `show_hidden` be `true` for admins, but it should not be forced.

In terms of code, I think this means that you have to switch from the `rest_validate_request_arg()` validation callback to a custom one. Or, if we don't like the idea of mixing responsibilities (validation vs calculating default values), perhaps we switch this param to `enum => array( true, false, null )` and then do a `if ( null === )` check in `get_items()` (the latter more closely matches what we do in BP core, but I'm not sure if it's idiomatic for the API endpoints).

@renatonascalves Do you have thoughts about the best way forward?

renatonascalves:
fixes #107

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -102,11 +102,6 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 			'page'               => $request['page'],
 		);
 
-		// Admins and moderators can see it all.
-		if ( bp_current_user_can( 'bp_moderate' ) ) {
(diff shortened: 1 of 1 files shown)


boonebgorges:
Yes, this pattern seems OK to me.

Issue #29, opened 2016-07-07
Simplify BuddyPress Namespace


renatonascalves:
The BuddyPress namespace should be simplified from buddypress/v1 into bp/v1.

This is how WordPress and WooCommece do theirs.

WordPress: wp/v2
WooCommerce: wc/v1

paulgibbs:
I don't know about "should". The fewer letters in the endpoint name increases the risk of conflicts if other plugins choose the same... this is probably unlikely, but that's why the full-length title was chosen.

I am not particularly worried if we decide to shorten it, however.

renatonascalves:
Definitely BuddyPress should take precedence over any plugin. But I'm gonna check how those guys, WC and WP, are checking for possible namespace conflicts...

modemlooper:
API endpoints need to be self explanatory. Naming should be full text. This will make sure a user knows its buddypress and as Paul explained will eliminate collisions.

renatonascalves:
I'm not sold, but I go where the consensus goes. =)

r-a-y:
I personally prefer the shortform as well.  Regardless, whatever we choose should be filterable so devs can change it.

modemlooper:
Actually, namespaces should be filterable. So, you should be able to change it to anything

renatonascalves:
Making a note I'll add a filter for the namespace. :)

rahulkhosla:
Please, could you update the docs? First time using this API and I didn't know that it was /wp-json/buddypress/v1/

renatonascalves:
@rahulkhosla Sure! I'll add more doc and some examples.

PR #114, opened 2018-12-13
Don't modify core /users/ query results.

boonebgorges:
The `bp_remove_has_published_posts_from_wp_api_user_query()` filter affects *all* API queries. This feels like unexpected behavior to me. IMO we should be more conservative here, only modifying the query if we know it's the BP `/members/` query.

First question is whether you agree with me about this. I don't have a really great argument :-D But, if I were building an external client that fetched a list of users at `/users/`, I'd expect to be able to use the `has_published_posts` param whether or not BP was involved.

Second question is how to implement. In this PR I specified the `/members/` endpoint, but if we extend the core `users` query elsewhere in BP-REST, then we might want to make it more general. The hardcoding of the namespace, version number, and `rest_base` also seem wrong to me - we might want these defined in a central location, both to be DRY and also to make them filterable (for whitelabeling, etc). See eg https://github.com/buddypress/BP-REST/issues/29#issuecomment-234601617

### DIFF
--- a/includes/bp-members/bp-members-filters.php
+++ b/includes/bp-members/bp-members-filters.php
@@ -22,6 +22,10 @@ defined( 'ABSPATH' ) || exit;
  * @return array
  */
 function bp_remove_has_published_posts_from_wp_api_user_query( $prepared_args, $request ) {
+	if ( 0 !== strpos( $request->get_route(), '/buddypress/v1/members' ) ) {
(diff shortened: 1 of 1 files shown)


renatonascalves:
The logic seems correct to me. I'll merge this. The endpoints abstraction will be done on #29.

Issue #116, opened 2018-12-13
More transparent parameters for activity queries?


boonebgorges:
Currently, to fetch activity related to a specific group, you have to do something like this: `GET /wp-json/buddypress/v1/activity?component=groups&primary_item_id=74`. In other words, you have to know that the BP activity schema, in the case of groups, stores 'groups' for `component` and the group ID for the `item_id`. And if you're querying for a different kind of activity, you have to know the schema for that as well - eg `GET /wp-json/buddypress/v1/activity?component=blogs&primary_item_id=123`, where `123` is the site ID.

This pattern closely mirrors not only our database schema, but also our PHP API, such as `bp_activity_get()`. Yet it seems to me that it requires too much knowledge of BuddyPress internals for an externally-facing API. Builders of client apps shouldn't need to know this much about how BP stores and queries for activity data.

My thought is that we might want to offer - alongside the existing `primary_item_id`, etc, which should continue to be supported - some parameters that are more straightforward for clients. Eg:
- `GET /wp-json/buddypress/v1/activity?site_id=123&type=new_blog_post`
- `GET /wp-json/buddypress/v1/activity?group_id=123`
etc. Internally, we'd translate into the proper `component` + `item_id`.

On the flip side, we would then offer the same translated info when querying for activity items. For example, a group-linked activity item currently looks like:

```json
 {
        "_links": {
           ...
        },
        "component": "groups",
        "content": {
            "rendered": "<p>abcde</p>\n"
        },
        "date": "2018-07-20T02:44:19",
        "favorited": null,
        "id": 244317,
        "link": "http://example.coml/wpmaster/activity/p/244317/",
        "parent": 0,
        "primary_item_id": 74,
        "secondary_item_id": 0,
        "status": "published",
        "title": "...",
        "type": "activity_update",
        "user": 1,
        "user_avatar": {
           ...
        }
    }
```

But we could easily add a `group_id` or `group` property, along with the appropriate item in `links`.

Thoughts?

PR #124, opened 2019-01-20
Group Membership Endpoint

### LINKED ISSUE #113: Group members endpoint
Looks like there's currently no way to handle group membership CRUD through the API.

My thought is that it probably makes sense to have this before a BP merge, at least if we plan to merge the other group-related endpoints. It feels to me like most use cases of the API (say, building a front-end interface) would need group membership functionality - at the very least, read-only to make the groups component useful.

renatonascalves:
fixes #113

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
@@ -40,7 +40,7 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 	 */
 	public function __construct() {
 		$this->namespace        = 'buddypress/v1';
-		$this->rest_base        = '/group/members';
+		$this->rest_base        = 'group/members';
 		$this->groups_endpoint  = new BP_REST_Groups_Endpoint();
 		$this->members_endpoint = new BP_REST_Members_Endpoint();
 	}
--- a/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
@@ -16,13 +16,23 @@ defined( 'ABSPATH' ) || exit;
 class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 
 	/**
+	 * Reuse some parts of the BP_REST_Groups_Endpoint class.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param object BP_REST_Groups_Endpoint
+	 */
+	protected $groups_endpoint;
+
+	/**
 	 * Constructor.
 	 *
 	 * @since 0.1.0
 	 */
 	public function __construct() {
-		$this->namespace = 'buddypress/v1';
-		$this->rest_base = '/group/members';
+		$this->namespace       = 'buddypress/v1';
+		$this->rest_base       = '/group/members';
+		$this->groups_endpoint = new BP_REST_Groups_Endpoint();
 	}
 
 	/**
@@ -57,7 +67,7 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 	 * @return WP_REST_Request List of group members.
 	 */
 	public function get_items( $request ) {
-		$group = $this->get_group_object( $request['group_id'] );
+		$group = $this->groups_endpoint->get_group_object( $request['group_id'] );
 
 		if ( ! $group ) {
 			return new WP_Error( 'bp_rest_invalid_group_id',
@@ -79,6 +89,10 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 			'exclude_admins_mods' => (bool) $request['exclude_admins'],
 			'exclude_banned'      => (bool) $request['exclude_banned'],
 		);
+
+		if ( empty( $args['exclude'] ) ) {
+			$args['exclude'] = false;
+		}
 
 		/**
 		 * Filter the query arguments for the request.
@@ -124,10 +138,10 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return bool
+	 * @return bool|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
-		return true;
+		return $this->groups_endpoint->get_item_permissions_check( $request );
 	}
 
 	/**
@@ -151,7 +165,7 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 			);
 		}
 
-		$group = $this->get_group_object( $request['group_id'] );
+		$group = $this->groups_endpoint->get_group_object( $request['group_id'] );
 
 		if ( ! $group ) {
 			return new WP_Error( 'bp_rest_invalid_group_id',
@@ -353,28 +367,6 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 	}
 
 	/**
-	 * Get group object from its identifier.
-	 *
-	 * @since 0.1.0
-	 *
-	 * @param int $group_id Group ID.
-	 * @return bool|BP_Groups_Group
-	 */
-	protected function get_group_object( $group_id ) {
-
-		// Get group object.
-		$group_obj = groups_get_group( array(
-			'group_id' => $group_id,
-		) );
-
-		if ( empty( $group_obj->id ) ) {
-			return false;
-		}
-
-		return $group_obj;
-	}
-
-	/**
 	 * Get the group member schema, conforming to JSON Schema.
 	 *
 	 * @since 0.1.0
@@ -411,14 +403,15 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 			'default'           => 'last_joined',
 			'type'              => 'string',
 			'enum'              => array( 'last_joined', 'first_joined' ),
-			'sanitize_callback' => 'rest_validate_request_arg',
+			'sanitize_callback' => 'sanitize_key',
+			'validate_callback' => 'rest_validate_request_arg',
 		);
 
 		$params['roles'] = array(
 			'description'       => __( 'Ensure result set includes specific group roles.', 'buddypress' ),
 			'default'           => array(),
 			'type'              => 'array',
-			'sanitize_callback' => 'wp_parse_id_list',
+			'sanitize_callback' => 'bp_rest_sanitize_string_list',
 		);
 
 		$params['search'] = array(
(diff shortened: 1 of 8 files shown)


includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+			'description'       => __( 'Offset the result set by a specific number of pages of results.', 'buddypress' ),
+			'default'           => 1,
+			'type'              => 'integer',
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+
+		$params['exclude'] = array(
+			'description'       => __( 'Ensure result set excludes specific member IDs.', 'buddypress' ),
+			'default'           => array(),
```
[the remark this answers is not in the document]
boonebgorges: A quick follow-up that I had to fix this before I could properly test. I added the following in `get_items()`:

```php

		if ( empty( $args['exclude'] ) ) {
			$args['exclude'] = false;
		}

```

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+			'type'              => 'array',
+			'sanitize_callback' => 'wp_parse_id_list',
+		);
+
+		$params['exclude_admins'] = array(
+			'description'       => __( 'Whether results should exclude group admins and mods.', 'buddypress' ),
+			'default'           => true,
```
boonebgorges: I wonder what you think about changing this default behavior. It's always confusing to me in BP that you don't get admins and mods by default. Third-party consumers will likely find it confusing as well. (Similar to #116, I think we should assume that clients have no prior knowledge of BP's PHP APIs, and we should take this opportunity to make the interface clearer.)

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+			'type'              => 'string',
+			'enum'              => array( 'last_joined', 'first_joined' ),
+			'sanitize_callback' => 'rest_validate_request_arg',
+		);
+
+		$params['roles'] = array(
+			'description'       => __( 'Ensure result set includes specific group roles.', 'buddypress' ),
+			'default'           => array(),
+			'type'              => 'array',
+			'sanitize_callback' => 'wp_parse_id_list',
```
boonebgorges: I believe that the roles are strings, not ints.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 0.1.0
+	 */
+	public function __construct() {
+		$this->namespace = 'buddypress/v1';
+		$this->rest_base = '/group/members';
```
boonebgorges: `/group/members` or `/groups/members/` ?

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+	/**
+	 * Check if a given request has access to group members.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return bool
+	 */
+	public function get_items_permissions_check( $request ) {
+		return true;
```
boonebgorges: This should respect the privacy level of groups. Is there a way to inherit `can_see()` from the groups controller?

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+
+	/**
+	 * Check if a given request has access to update a group.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function update_item_permissions_check( $request ) {
```
boonebgorges: This will need to cover some more cases:

1. Groups admins can remove/ban/promote users
2. Users can remove themselves from groups
3. Users can add themselves to public groups
4. Users can request memberships to private groups (though I'm unsure whether this should be handled here or not)

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return array
+	 */
+	public function get_update_collection_params() {
+		$params                       = parent::get_collection_params();
+		$params['context']['default'] = 'view';
+
+		$params['id'] = array(
```
boonebgorges: IMO this should be `user_id`. Group memberships have unique IDs that are distinct from the user IDs, and calling this param `id` could cause confusion. It also is somewhat unclear given that we need to pass `group_id`.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+		if ( (int) $id <= 0 ) {
+			return false;
+		}
+
+		$user = get_userdata( (int) $id );
+		if ( empty( $user ) || ! $user->exists() ) {
+			return false;
+		}
+
+		if ( is_multisite() && ! is_user_member_of_blog( $user->ID ) ) {
```
boonebgorges: I think that, on many multisite setups, BP users do not have a role on the main site.

tests/groups/test-group-member-controller.php
```diff
@@ -0,0 +1,334 @@
...
+
+		$request->set_param( 'context', 'view' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 200, $response->get_status() );
+
+		$all_data = $response->get_data();
+		$data     = $all_data;
+
+		foreach ( $all_data as $data ) {
```
boonebgorges: These tests are all passing locally for me even though they're returning no members (see my comment about `exclude` above) and I think it's because the `foreach` loop isn't triggered when `$all_data` is empty. Putting `assertNotEmpty()` before is probably sufficient.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+
+		$params['status'] = array(
+			'description'       => __( 'Sort the order of results by the status of the group members.', 'buddypress' ),
+			'default'           => 'last_joined',
+			'type'              => 'string',
+			'enum'              => array( 'last_joined', 'first_joined' ),
+			'sanitize_callback' => 'rest_validate_request_arg',
```
boonebgorges: This should be `validate_callback`. This typo is causing tests to fail, because the `sanitize_callback` is causing a chain-reaction that's triggering a `type=online` query, which always returns no members during unit testing :-D

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
renatonascalves: @boonebgorges I think `/group/members`, singular, as we are fetching members of a specific, single, group instead of members of several groups.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
renatonascalves: No caps, no nothing? `is_user_member_of_blog` doesn't check roles, but caps.

boonebgorges: Correct. It looks like `bp_members_add_role_after_activation()` was added in 3.0.0 to fix this, but prior to this (and so affecting many older installations) it's possible to be a BP member without being a main-site user. See https://buddypress.trac.wordpress.org/ticket/7565.

I'd lean toward removing this check.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
boonebgorges: Gotcha. Neither one seems great to me, to be honest :) When read in a RESTful way, `group/members/` implies that there's a group called `members`. What do you think of removing the (pseudo-)nesting - `$rest_base = '/group-members'` ?

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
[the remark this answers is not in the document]
boonebgorges: I've opened a BP ticket describing the issue: https://buddypress.trac.wordpress.org/ticket/8040#ticket Up to you whether you want to change the behavior here too.

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -0,0 +1,517 @@
...
+
+	/**
+	 * Get the group member schema, conforming to JSON Schema.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return array
+	 */
+	public function get_item_schema() {
+		$members_endpoint = new BP_REST_Members_Endpoint();
```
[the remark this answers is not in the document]
renatonascalves: Added! :)

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
renatonascalves: Got it! Removing it! :)

includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
renatonascalves: @boonebgorges 1, 2 and 3 were added. Number 4 will be tackled later!

boonebgorges:
Thanks, @renatonascalves ! Your logic was mostly right. I added a clause ensuring that users are not able to add other users to public groups.

I found it quite hard to follow the permission checks the way you'd written them, so I had a go at reorganizing. I put the `bp_moderate` check first. Then, I handled self-requests (user_id === loggedin user) separately from requests about other users. I think this simplifies the flow.

Note that, in the `default` case, I'm just returning `false`. This throws the regular `rest_forbidden` error instead of a custom one. I don't feel strongly about this - it makes the logic much simpler not to have a separate error for each failure, but it also provides less info for the client. Also, I noticed you're serving 500s for these custom errors, but the core endpoints use `rest_authorization_required_code()`, which is `403`. Is there a reason for this?

PR #125, opened 2019-01-21
Transparent Params for Activity Queries

### LINKED ISSUE #116: More transparent parameters for activity queries?

renatonascalves:
fixes #116

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -112,10 +112,8 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			'sort'              => $request['order'],
 			'spam'              => $request['status'],
 			'display_comments'  => $request['display_comments'],
-
 			'site_id'           => $request['site_id'],
 			'group_id'          => $request['group_id'],
-
 			'count_total'       => true,
 			'fields'             => 'all',
 			'show_hidden'       => false,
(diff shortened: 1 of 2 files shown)


renatonascalves:
@boonebgorges What do you think of this approach?

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -127,19 +129,31 @@ public function get_items( $request ) {
...
+			$args['filter']['object']     = 'blogs';
+			$args['filter']['primary_id'] = $args['site_id'];
 		}
 
-		$item_id = 0;
-		if ( ! empty( $request['primary_id'] ) ) {
-			$primary_id                   = $request['primary_id'];
-			$args['filter']['primary_id'] = $primary_id;
-			$item_id                      = $primary_id[0];
+		if ( empty( $args['group_id'] ) || empty( $args['site_id'] ) ) {
```
boonebgorges: should this be `&&` ? then it basically becomes an `else`

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -963,6 +977,15 @@ protected function prepare_links( $activity ) {
 			);
 		}
 
+		if ( 'groups' === $activity->component && ! empty( $activity->item_id ) ) {
+			$group = groups_get_group( $activity->item_id );
+
+			$links['group'] = array(
```
boonebgorges: <3

tests/activity/test-controller.php
```diff
@@ -110,6 +110,48 @@ public function test_get_public_groups_items() {
...
+			'group_id' => $g2,
+		) );
+
+		$request->set_param( 'context', 'view' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 200, $response->get_status() );
+		$a_ids = wp_list_pluck( $response->get_data(), 'id' );
+
+		$this->assertContains( $a2, $a_ids );
```
boonebgorges: As written, this test will pass without the param changes in the controller; since all activity items are linked to `$g2`, they're all returned whether we limit to `$g2` or not. We should add an `$a3` that is *not* linked to the group, and then either `assertEqualSets()` for `[ $a1, $a2 ]`, or `assertNotContains( $a3, $a_ids )`.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
renatonascalves: Ops! Fixing it!

tests/activity/test-controller.php
renatonascalves: Sorry! Updated!

renatonascalves:
@boonebgorges Updated with the suggestions! =)

PR #128, opened 2019-01-25
XProfile Data Endpoints

### LINKED ISSUE #122: Update user xprofile data endpoint
It looks like there's no way to update a user's xprofile data at the moment. Is that right, or am I just missing it? We should have it :) The question is whether it would be part of the `/members/` endpoint, or whether it should be part of `/xprofile/`.

renatonascalves:
fixes #122


renatonascalves:
@boonebgorges Could you code review this PR? :)

boonebgorges:
Yup, added to my list.

PR #126, opened 2019-01-21
Group Invitation Endpoints

renatonascalves:
Implementation of the group invitations.

This is ready for code review.

PR #135, opened 2019-01-27
Updated URI format for group member endpoint.

boonebgorges:
I began a review of #126 and, as in the case of the group membership endpoint, I found myself unhappy with the semantics of the URI format. `/group/invites/123`, where 123 is the user ID, doesn't make much sense to me.

So I thought I'd suggest a different format. Instead of applying it to the WIP invites endpoint, I've applied it to the already-merged members endpoint. Here's the new format:

`GET /groups/<group-id>/members`
`PUT (or POST) /groups/<group-id>/members/<user-id>`
`DELETE /groups/<group-id>/members/<user-id>`

This feels much more semantic, and I think we can apply a similar pattern elsewhere in the endpoints (such as for group invites). @renatonascalves Can you please give me your opinion on whether this is an improvement over what's currently there?

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
@@ -388,7 +388,7 @@ class BP_REST_Group_Members_Endpoint extends WP_REST_Controller {
 	public function delete_item_permissions_check( $request ) {
 		if ( ! is_user_logged_in() ) {
 			return new WP_Error( 'bp_rest_authorization_required',
-				__( 'Sorry, you need to be logged in to make an update.', 'buddypress' ),
+				__( 'Sorry, you need to be logged in to delete a group membership.', 'buddypress' ),
 				array(
 					'status' => rest_authorization_required_code(),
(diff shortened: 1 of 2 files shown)


includes/bp-groups/classes/class-bp-rest-group-members-endpoint.php
```diff
@@ -335,6 +329,130 @@ public function update_item_permissions_check( $request ) {
...
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|bool
+	 */
+	public function delete_item_permissions_check( $request ) {
+		if ( ! is_user_logged_in() ) {
+			return new WP_Error( 'bp_rest_authorization_required',
+				__( 'Sorry, you need to be logged in to make an update.', 'buddypress' ),
```
renatonascalves: This message should say `to delete a group membership`.

boonebgorges: Good catch, thanks.

PR #143, opened 2019-02-03
Make sure xProfile component is active before using it

imath:
The members endpoint, when using `bp_xprofile_get_groups()`, should check if the `xprofile` component is active first. Otherwise it will generate a fatal error if the xProfile component is not active.


imath:
Ok, why not ! see 88ff2c9

PR #144, opened 2019-02-03
Make sure bp_remove_has_published_posts_from_wp_api_user_query is doi…

imath:
…ng something :)

By removing leading and trailing slashes from the request route.

For instance my requests was `'/buddypress/v1/members'` and as a not logged in user I could only see the users who published a post.


renatonascalves:
@imath Just a notice this probably will be removed on #145

PR #159, opened 2019-02-09
Endpoints Permissions Check Override

### LINKED ISSUE #158: Permissions Check Override
We should add a filter in the permission checks so that anyone can override or add their own verification before doing ours.

An interesting scenario I came across recently was validating a token. Right now the API doesn't allow to add other verifications on top of what's already there. Forcing the API user to duplicate code extending the class to override it. So a filter would facilitate that job.

My plan is to add filters with the suggested pattern:

`bp_rest_COMPONENT_NAME_API_ACTION_permissions_check`

For example, the `delete_item` endpoint of the group component would be something like:

`bp_rest_groups_delete_item_permissions_check`

renatonascalves:
fixes #158

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -222,25 +222,19 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full data about the request.
-	 * @return bool
+	 * @return bool|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
 
 		/**
-		 * Filter or override the activity `get_items` permissions check.
+		 * Filter the activity `get_items` permissions check.
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param bool            $retval  Returned valued. Default: Always `true`.
+		 * @param bool|WP_Error   $retval  Returned value.
 		 * @param WP_REST_Request $request The request sent to the API.
 		 */
-		$retval = apply_filters( 'bp_rest_activity_get_items_permissions_check', true, $request );
-
-		if ( is_wp_error( $retval ) || ! $retval ) {
-			return $retval;
-		}
-
-		return (bool) $retval;
+		return apply_filters( 'bp_rest_activity_get_items_permissions_check', true, $request );
 	}
 
 	/**
@@ -285,23 +279,10 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @return bool|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-
-		/**
-		 * Filter or override the activity `get_item` permissions check.
-		 *
-		 * @since 0.1.0
-		 *
-		 * @param bool            $retval  Returned valued. Default: Always `true`.
-		 * @param WP_REST_Request $request The request sent to the API.
-		 */
-		$retval = apply_filters( 'bp_rest_activity_get_item_permissions_check', true, $request );
-
-		if ( is_wp_error( $retval ) || ! $retval ) {
-			return $retval;
-		}
+		$retval = true;
 
 		if ( ! $this->can_see( $request ) ) {
-			return new WP_Error( 'bp_rest_user_cannot_view_activity',
+			$retval = new WP_Error( 'bp_rest_user_cannot_view_activity',
 				__( 'Sorry, you cannot view the activities.', 'buddypress' ),
 				array(
 					'status' => rest_authorization_required_code(),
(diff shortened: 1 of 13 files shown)


includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -477,7 +537,7 @@ protected function can_see( $group_id ) {
 		 * @param int  $user_id   User ID.
 		 * @param int  $group_id  Group ID.
 		 */
-		return (bool) apply_filters( 'bp_rest_group_invite_can_see', $retval, $user_id, $group_id );
+		return (bool) apply_filters( 'bp_rest_group_invites_can_see', $retval, $user_id, $group_id );
```
boonebgorges: What's the relationship between the new endpoints-specific filters and the more general `can_see` filters? Should we get rid of the latter?

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -270,6 +285,21 @@ public function get_item( $request ) {
...
+
+		/**
+		 * Filter or override the activity `get_item` permissions check.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param bool            $retval  Returned valued. Default: Always `true`.
+		 * @param WP_REST_Request $request The request sent to the API.
+		 */
+		$retval = apply_filters( 'bp_rest_activity_get_item_permissions_check', true, $request );
```
boonebgorges: This pattern, used throughout the PR, allows you to use the filters to deny access to an endpoint that a user would usually have access to, by returning `false` or `WP_Error`. But it doesn't let you *grant* access to a filter that's normally off-limits, because (a) `true` is the default value, and (b) it runs before the additional checks that happen in the method. If we want this to be a general system for filtering permissions, it needs to come at the end of the method, and we need to avoid early `return` in the method. So:

```php
$retval = true;

if ( ! $this->can_see( $request ) ) {
    $retval = new WP_Error...
}

...

return apply_filters( ... )
```

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
renatonascalves: @boonebgorges I'll remove those. They made sense before but now I don't think they are needed.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
renatonascalves: @boonebgorges Makes sense. The scenario I envisioned was doing a token verification, and user validation, before the other permissions checks were made.

But I just realized this particular scenario has a flaw as the token verification should be added in the header of the request and not in the request `params`, something the client is doing.

See this example:

```
$AP_JWT = new AP_JWT();

$validation = $AP_JWT->validate_token( $request );

// Bail early.
if ( $validation !== true ) {

    if( !empty( $validation['error'] ) ) {
        $msg = $validation['error'];
    }
    return new WP_Error( 'rest_authorization_required',
        __( 'Sorry, you are not allowed to do that. ' . $msg, 'text-domain' ),
        array(
            'status' => rest_authorization_required_code(),
        )
    );
}

Then the other permission checks were done here.
```

In the current BP implementation, you can't verify this token FIRST beforing the other permissions check. With your suggestion, this scenario wouldn't work because the user validation would be done after and it would be returning an error, as the user was non-existent before.

I see this is not a BuddyPress problem, it is just a problem on how this particular client is verifying a request.

So I'll go ahead with your approach, as it takes into account a broad use case rather this specific one. =D

renatonascalves: @boonebgorges Another problem is that the order of the permissions check was important. You could not test `bp_activity_user_can_delete` if the activity didn't exist.

With this process, that's not possible. :/

boonebgorges: @renatonascalves Yes, it may take some rethinking about how the checks are organized. Here's one way of thinking about it:

```php
	public function update_item_permissions_check( $request ) {
		$retval = true;

		if ( ! is_user_logged_in() ) {
			$retval = new WP_Error( 'bp_rest_authorization_required',
				__( 'Sorry, you are not allowed to update this activity.', 'buddypress' ),
				array(
					'status' => rest_authorization_required_code(),
				)
			);
		}

		$activity = $this->get_activity_object( $request );

		if ( true === $retval && empty( $activity->id ) ) {
			$retval = new WP_Error( 'bp_rest_activity_invalid_id',
				__( 'Invalid activity id.', 'buddypress' ),
				array(
					'status' => 404,
				)
			);
		}

		if ( true === $retval && ! bp_activity_user_can_delete( $activity ) ) {
			return new WP_Error( 'bp_rest_activity_cannot_update',
				__( 'Sorry, you are not allowed to update this activity.', 'buddypress' ),
				array(
					'status' => rest_authorization_required_code(),
				)
			);
		}

		/**
		 * Filter or override the activity `update_item` permissions check.
		 *
		 * @since 0.1.0
		 *
		 * @param bool            $retval  Returned valued. Default: Always `true`.
		 * @param WP_REST_Request $request The request sent to the API.
		 */
		$retval = apply_filters( 'bp_rest_activity_update_item_permissions_check', true, $request );


		return $retval;
	}

```

boonebgorges: (in other words, if your first check fails, you don't bother to do the subsequent checks)

renatonascalves: Cool! =)

PR #161, opened 2019-02-11
Avatar Upload Endpoints

### LINKED ISSUE #154: Avatar Upload Endpoint
This issue is to keep track of adding support to add avatar upload.

- `create_item` - Add image.
- `delete_item` - Delete image.
- `get_item` - Get avatar (?)

renatonascalves:
fixes #154


renatonascalves:
@boonebgorges  The avatar endpoint is ready for another round of code review.

Here the most important changes, based on your last feedback:

• deletion of previous/existing avatar added.
• `thumb` and `full` avatars are shrunk to their proper proportions.
• Avatars with incorrect aspect ratio are not auto-cropped based on your suggestion. 

This last part is the one I really would like an opinion on. I tested uploading via the endpoint with two to three images with really different sizes and the way I did it, based on how BP core is doing it, is not consistent or at least, really weird. 

Let me know what you think. :)

boonebgorges:
Nice! This is working pretty well in my tests. For reference, I'm using https://httpie.org/doc as follows:

```
http -f POST path/to/wp-json/buddypress/v1/members/2/avatar 'Cookie:wordpress_logged_in_174edc99138352bc8dd30c9ca2344d88=boone1|thecookievalue' X-WP-Nonce:thenonce file@/path/to/file.jpg
```

Deletion and full/thumb resizing seem to be working well.

Auto-cropping is indeed weird. You copied the logic that's used in BP to set the default dimensions of the cropper. But this is too conservative for the dimensions of the auto-cropper - we should be using as much of the image as possible, instead of as little as possible. I've just pushed up some improved logic, which seems to work much better in my tests. Could you have a look?

In an ideal world, we might allow the endpoint to accept `crop_x`, `crop_w`, `crop_y`, and `crop_h` params. That being said, third-party apps are likely to have their own cropper logic, and to send images to the endpoint that already have the correct aspect ratio. So I don't think that this additional logic is a blocker.

PR #163, opened 2019-02-17
Request a Group Membership

### LINKED ISSUE #162: Private Group Request

renatonascalves:
fixes #162

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -190,26 +190,45 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 		$group_id     = $group->id;
 		$group_member = new BP_Groups_Member( $user->ID, $group_id );
 
-		// Add member to the group.
-		$group_member->group_id     = $group_id;
-		$group_member->user_id      = $user->ID;
-		$group_member->is_admin     = 0;
-		$group_member->date_modified = bp_core_current_time();
-		$group_member->is_confirmed  = 1;
-		$saved                      = $group_member->save();
+		// Membership request to private group.
+		if ( 'private' === $group->status ) {
+			$membership_request = groups_send_membership_request( $user->ID, $group_id );
 
-		if ( ! $saved ) {
-			return new WP_Error( 'bp_rest_group_member_failed_to_join',
-				__( 'Could not add member to the group.', 'buddypress' ),
-				array(
-					'status' => 500,
-				)
-			);
-		}
+			if ( ! $membership_request ) {
+				return new WP_Error( 'bp_rest_group_membership_request_failed',
+					__( 'Could not send membership request to this group.', 'buddypress' ),
+					array(
+						'status' => rest_authorization_required_code(),
+					)
+				);
+			}
 
-		// If new role set, promote it too.
-		if ( $saved && 'member' !== $role ) {
-			groups_promote_member( $user->ID, $group_id, $role );
+			// Redifing here to fetch updated values.
+			$group_member = new BP_Groups_Member( $user->ID, $group_id );
+
+		} else {
+
+			// Add member to the group.
+			$group_member->group_id      = $group_id;
+			$group_member->user_id       = $user->ID;
+			$group_member->is_admin      = 0;
+			$group_member->date_modified = bp_core_current_time();
+			$group_member->is_confirmed  = 1;
+			$saved                       = $group_member->save();
+
+			if ( ! $saved ) {
+				return new WP_Error( 'bp_rest_group_member_failed_to_join',
+					__( 'Could not add member to the group.', 'buddypress' ),
+					array(
+						'status' => 500,
+					)
+				);
+			}
+
+			// If new role set, promote it too.
+			if ( $saved && 'member' !== $role ) {
+				groups_promote_member( $user->ID, $group_id, $role );
+			}
 		}
 
 		$retval = array(
@@ -221,7 +240,7 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 		$response = rest_ensure_response( $retval );
 
 		/**
-		 * Fires after a member us added to a group via the REST API.
+		 * Fires after a member is added, or a membership request, to a group via the REST API.
 		 *
 		 * @since 0.1.0
 		 *
@@ -286,7 +305,12 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 			$loggedin_user_id = bp_loggedin_user_id();
 			if ( $loggedin_user_id === $user->ID ) {
 
-				// Users may only freely join public groups. @todo Private requests.
+				// create a group membership request.
+				if ( true === $retval && 'private' === $group->status ) {
+					return true;
+				}
+
+				// Users may only freely join public groups.
 				if ( true === $retval && 'public' !== $group->status && ! groups_is_user_member( $loggedin_user_id, $group->id ) ) {
 					$retval = new WP_Error( 'bp_rest_group_member_cannot_join',
 						__( 'Sorry, you are not allowed to join this group.', 'buddypress' ),
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -190,45 +190,26 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 		$group_id     = $group->id;
 		$group_member = new BP_Groups_Member( $user->ID, $group_id );
 
-		// Membership request to private group.
-		if ( 'private' === $group->status ) {
-			$membership_request = groups_send_membership_request( $user->ID, $group_id );
+		// Add member to the group.
+		$group_member->group_id      = $group_id;
+		$group_member->user_id       = $user->ID;
+		$group_member->is_admin      = 0;
+		$group_member->date_modified = bp_core_current_time();
+		$group_member->is_confirmed  = 1;
+		$saved                       = $group_member->save();
 
-			if ( ! $membership_request ) {
-				return new WP_Error( 'bp_rest_group_membership_request_failed',
-					__( 'Could not send membership request to this group.', 'buddypress' ),
-					array(
-						'status' => rest_authorization_required_code(),
-					)
-				);
-			}
+		if ( ! $saved ) {
+			return new WP_Error( 'bp_rest_group_member_failed_to_join',
+				__( 'Could not add member to the group.', 'buddypress' ),
+				array(
+					'status' => 500,
+				)
+			);
+		}
 
-			// Redifing here to fetch updated values.
-			$group_member = new BP_Groups_Member( $user->ID, $group_id );
-
-		} else {
-
-			// Add member to the group.
-			$group_member->group_id      = $group_id;
-			$group_member->user_id       = $user->ID;
-			$group_member->is_admin      = 0;
-			$group_member->date_modified = bp_core_current_time();
-			$group_member->is_confirmed  = 1;
-			$saved                       = $group_member->save();
-
-			if ( ! $saved ) {
-				return new WP_Error( 'bp_rest_group_member_failed_to_join',
-					__( 'Could not add member to the group.', 'buddypress' ),
-					array(
-						'status' => 500,
-					)
-				);
-			}
-
-			// If new role set, promote it too.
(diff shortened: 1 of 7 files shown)


includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -190,26 +190,45 @@ public function create_item( $request ) {
...
-			return new WP_Error( 'bp_rest_group_member_failed_to_join',
-				__( 'Could not add member to the group.', 'buddypress' ),
-				array(
-					'status' => 500,
-				)
-			);
-		}
+		// Membership request to private group.
+		if ( 'private' === $group->status ) {
+			$membership_request = groups_send_membership_request( $user->ID, $group_id );
```
renatonascalves: `groups_send_membership_request` already does several permissions checks inside it. So rather than re-adding them in this endpoint, I decided to use it.

The only drawback I can see is lack of information of type of error: meaning the user won't know why the request won't go through.

boonebgorges: I think this is fine for now. In the future, we can improve this by abstracting the permission check out of `group_send_membership_request()`.

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -190,26 +190,45 @@ public function create_item( $request ) {
...
+			if ( ! $membership_request ) {
+				return new WP_Error( 'bp_rest_group_membership_request_failed',
+					__( 'Could not send membership request to this group.', 'buddypress' ),
+					array(
+						'status' => rest_authorization_required_code(),
+					)
+				);
+			}
+
+			// Redifing here to fetch updated values.
```
boonebgorges: spelling :-D `Redefining`

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -286,7 +305,12 @@ public function create_item_permissions_check( $request ) {
 			$loggedin_user_id = bp_loggedin_user_id();
 			if ( $loggedin_user_id === $user->ID ) {
 
-				// Users may only freely join public groups. @todo Private requests.
+				// create a group membership request.
```
boonebgorges: This check suggests that perhaps we need to change the semantics. Imagine you have three groups: `A` is public, `B` is private, `C` is hidden. As you've written it, `POST groups/A/members` will result in a group join, as will `POST groups/C/members` (assuming the proper permissions), while `POST groups/B/members` will result in a request. This asymmetry feels correct. It also means that it's impossible to add users directly to a private group using the API, which is something that admins ought to be able to do (in the admin panel, for instance).

Two alternatives:
1. A separate `membership-request` endpoint. So `POST /groups/B/membership-request`. This would also give us a convenient way of processing `accept` and `reject` actions, which I don't think are reflected here yet.
2. A `force` param which would allow properly-authorized requests to skip the request for private groups, doing an add instead. (This only fixes the second of the problems I describe above.)

imath:
Hi @renatonascalves & @boonebgorges 

Thanks a lot for your work/review. I'll need an extra delay to add mine, sorry 😞. This wednesday for sure.

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
renatonascalves: @boonebgorges I'd rather do it right the first time than something that might need an update later on.

I'm just not convinced it should go to a different endpoint. The private group scenario is just one example that the endpoint could take care of internally.

boonebgorges: I understand not wanting to overcomplicate things. But IMO it's an accident of history that group membership requests (and invites) and group memberships are stored in the same way. A couple of points:
- From the point of view of a third-party client, they don't feel related at all. You may attempt to create a membership, but get back a 403 saying 'you must request membership to this group'; so then, you create a membership-request. 
- The verbs we use with memberships (create, ban, promote, remove) are different from what we use with requests (create, approve, reject). 
- There may be a future time when requests and invites (which IMO are very similar - think of the verbs) are not stored as group memberships at all. See https://buddypress.trac.wordpress.org/ticket/6210
- Furthermore, while it's currently the case in BP's front-end interfaces that you can only request membership to *private* groups, there's no reason that this needs to be baked into the API. I can certainly imagine building an app that would allow requests to hidden groups if you knew the URL; or that would force requests to be sent for public groups. For this reason, it makes sense not to hardcode requests as being linked to private groups.

In sum, memberships and requests just feel like *different kinds of things*. cc @dcavins who has thought a lot about this.

dcavins: Thanks for the ping, Boone. Yes, I'm currently at work on a new patch for the long-dormant 6210. The change will generalize requests and invitations, so they can be applied to different kinds of objects (groups, sites, networks, or even forums or BP Docs for example). Part of the change will necessitate moving invite and request storage out of the group table (which will be a good thing in my opinion). A request/invitation endpoint that is separate from the create membership endpoint makes sense to me, and seems more obvious for devs who don't have a deep understanding of how BP works.

I've also, in the past, dallied with making group statuses more dynamic, rather than public, private and hidden are what they are, so would love it if requests are not handled in a strict way based on the group status here. Please use the `user_can` check for a particular capability (which at the moment check the status, but most likely will not rely on status in the future): https://github.com/buddypress/BuddyPress/blob/master/src/bp-groups/bp-groups-filters.php#L194

Thanks!

boonebgorges: Ah, I didn't realize those caps had been implemented in BP! Glad I pinged you here @dcavins :)

dcavins: Thanks! 👍 Beetlejuice, Beetlejuice...

imath:
I still haven’t found the time to test, but reading @boonebgorges & @dcavins comments, I agree requests should probably have their own endpoint or be into the invites one (as users in both cases are not yet members of the group: they need to accept the invite or they need an admin to accept the request). I also have a concern regarding the WordPress Single Group Admin screens. As I’ve said I haven’t tested yet but looking quickly at the `private` group status check (which seems to force a membership request) I’m wondering if a SuperAdmin can still directly add a new member to a private group using the REST API.

imath
https://imathi.eu

________________________________
De : David Cavins <notifications@github.com>
Envoyé : mardi, février 19, 2019 4:10 AM
À : buddypress/BP-REST
Cc : imath; Mention
Objet : Re: [buddypress/BP-REST] Request to Private Group Membership (#163)


@dcavins commented on this pull request.

________________________________

In includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php<https://github.com/buddypress/BP-REST/pull/163#discussion_r257877571>:

> @@ -286,7 +305,12 @@ public function create_item_permissions_check( $request ) {

                        $loggedin_user_id = bp_loggedin_user_id();

                        if ( $loggedin_user_id === $user->ID ) {



-                               // Users may only freely join public groups. @todo Private requests.

+                               // create a group membership request.



Thanks! 👍 Beetlejuice, Beetlejuice...

renatonascalves:
Thank you @boonebgorges, @dcavins and @imath for your insights.

I will be updating the code add this custom endpoint at `membership-request`, using `bp_user_can`, something I didn't know it existed.

I do think this particular endpoint will be temporary, meaning in next versions I think a new place would be chosen/created.

For example, the work @dcavins is doing with the invitation api, https://buddypress.trac.wordpress.org/ticket/6210, seems really interesting. And the REST API will definetely make use of that.

:)

renatonascalves:
@imath, @boonebgorges, and @dcavins 

Guys! Code updated to allow `accept/reject` via params. I thought it was better than adding a `DELETE` endpoint. The reason was only semantics.

Also, I didn't use `bp_user_can` because it is attached to the `private` [group status](https://github.com/buddypress/BuddyPress/blob/4cb16c0c326e7e12209519b489856738f3b64564/src/bp-groups/bp-groups-filters.php#L252), and the current implementation allows groups with any status to receive membership requests.

Let me know what you guys think. :)

imath:
Thanks a lot for your work @renatonascalves I just gave it a quick look and haven’t food a get_items & get_item methods. How do you get membership requests for a group. Do we need to use the Group Memberships endpoint instead ?

renatonascalves:
@imath Good point! I didn't think it was necessary, but I'm happy to add that. :)

imath:
I think it would make sense yes. Also, I think it would be more accurate to use the `DELETE` verb to remove a membership request if it removes a line into the DB table. I don’t remember if it’s the case 😬

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -0,0 +1,572 @@
...
+					'status' => 500,
+				)
+			);
+		}
+
+		// Site administrators can do anything.
+		if ( true === $retval && bp_current_user_can( 'bp_moderate' ) ) {
+			$retval = true;
+		} else {
+			if ( true === $retval && ( ! groups_is_user_admin( $user->ID, $group->id ) || ! groups_is_user_mod( $user->ID, $group->id ) ) ) {
```
imath: Only group admins can manage membership requests as it's using the `bp_is_item_admin()` check. I think it's best to not allow group mods.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -0,0 +1,572 @@
...
+				)
+			);
+		}
+
+		// Site administrators can do anything.
+		if ( true === $retval && bp_current_user_can( 'bp_moderate' ) ) {
+			$retval = true;
+		} else {
+			$user_id = get_current_user_id();
+			if ( true === $retval && ( ! groups_is_user_admin( $user_id, $group->id ) || ! groups_is_user_mod( $user_id, $group->id ) ) ) {
```
imath: Only group admins can manage membership requests as it's using the `bp_is_item_admin()` check. I think it's best to not allow group mods.

renatonascalves:
@imath Updated with your suggestions and the `get_item` endpoint.

About the batch/update, this will be tackled later. There are several places this will be useful, but I decided to tackle this feature/problem after we have something stable and working.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -0,0 +1,669 @@
...
+
+	/**
+	 * Membership slug.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @var $membership_slug
+	 */
+	protected $membership_slug;
+
```
imath: 1. I think i'll add a protected membership property to avoid a dq query when getting the membership as it's already done into the `get_item_permissions_check()` method.

eg: `protected $membership;`

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -0,0 +1,669 @@
...
+			$retval = new WP_Error( 'bp_rest_authorization_required',
+				__( 'Sorry, you need to be logged in to get a membership.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		}
+
+		$user_id    = bp_loggedin_user_id();
+		$membership = new BP_Groups_Member( false, false, absint( $request['membership_id'] ) );
```
imath: 2. I would use the protected property set in point 1

eg: `$this->membership = new BP_Groups_Member( false, false, absint( $request['membership_id'] ) );`

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -0,0 +1,669 @@
...
+	 * Fetch pending group membership request.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  WP_REST_Request $request Full data about the request.
+	 * @return WP_REST_Request|WP_Error
+	 */
+	public function get_item( $request ) {
+		// Get membership.
+		$membership = new BP_Groups_Member( false, false, absint( $request['membership_id'] ) );
```
imath: 3. then i would use `$this->membership` directly removing this line and replacing `$membership` by `$this->membership` at line 241

renatonascalves:
@imath Good idea! Actually, this pattern can be reused in other components as well. Let's do this! =D

PR #166, opened 2019-03-09
Messages Endpoints

### LINKED ISSUE #153: Messages Endpoints
Right now it is only possible to list messages. This component needs better support:

- `create_item` - Create message and thread.
- `delete_item` - Delete message and thread.
- `get_item` - See message and thread.

renatonascalves:
fixes #153

### DIFF
--- a/includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
+++ b/includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
@@ -38,6 +38,25 @@ class BP_REST_Messages_Endpoint extends WP_REST_Controller {
 				'permission_callback' => array( $this, 'get_items_permissions_check' ),
 				'args'                => $this->get_collection_params(),
 			),
+			array(
+				'methods'             => WP_REST_Server::CREATABLE,
+				'callback'            => array( $this, 'create_item' ),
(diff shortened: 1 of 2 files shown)


includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
```diff
@@ -116,9 +136,9 @@ public function get_items_permissions_check( $request ) {
 			);
 		}
 
-		if ( true === $retval && ! $this->can_see() ) {
-			$retval = new WP_Error( 'bp_rest_user_cannot_view_messages',
-				__( 'Sorry, you cannot view the messages.', 'buddypress' ),
+		if ( true === $retval && ! bp_current_user_can( 'bp_moderate' ) ) {
```
imath: The user cannot see the messages he's concerned with 😭 This statement should probably look like this:
```php
$user = bp_rest_get_user( $request['user_id'] );
if ( true === $retval && (int) $user->ID !== bp_loggedin_user_id() && ! bp_current_user_can( 'bp_moderate' ) ) {
```

imath:
@renatonascalves I think this endpoint needs some more work and should probably wait for next milestone.

PR #175, opened 2019-03-23
PHPCS Support

renatonascalves:
And minor travis refactor.


renatonascalves:
that was fun to test! =P

PR #189, opened 2019-06-26
Implement additional fields for Activity, Groups, Members & Notificat…

imath:
…ions

Althought the `WP_REST_Controller::add_additional_fields_to_object` were used into these components, there was no way to actually get these fields or set them.
This PR do the necessary stuff:
- make sure the title property of the schema is `bp` prefixed and uses the component ID as this property is used to get the registered additional fields for the endpoint.
- wrap `register_rest_field` into a specific BP REST function making sure to prefix all registered field with `bp` as this global is used for custom post types
- make sure the endpoint schema can be appended with the custom fields schema
- make sure creating/updating an item triggers the `WP_REST_Controller::update_additional_fields_for_object` method

I also removed `WP_REST_Controller::add_additional_fields_to_object`:
1. from avatar endpoints (I am wondering what avatar meta could be used?),
2. from the Message endpoint as it is actually a Thread endpoint and because there is something wrong with cache,
3. from xprofiles because it needs more work as meta for xprofile can be typed for data, group or field. On this last point I suggest to discuss about the best scenario (adding a last parameter to `bp_rest_register_field` and calling the object type and schema title `bp_xpofile_$type` seems to be a good way to do it)

PS: About 2. & 3. I'll open GH issues soon.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -396,9 +396,16 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			)
 		);
 
+		$activity      = current( $activity['activities'] );
+		$fields_update = $this->update_additional_fields_for_object( $activity, $request );
+
+		if ( is_wp_error( $fields_update ) ) {
+			return $fields_update;
+		}
+
 		$retval = array(
 			$this->prepare_response_for_collection(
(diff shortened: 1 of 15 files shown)


imath:
Ouch phpcs issues I’ll update the PR to fix these 😅

imath:
Thanks a lot for giving it a look @boonebgorges I think we should go with it. This first step will allow me to work on xProfiles meta specific case from here #191

@renatonascalves The WPCS issue should be fixed with my latest commit on the PR, please commit it to master if you also feel the same way 😉

PR #196, opened 2019-07-24
Try/consistent method arguments

imath:
Writing the documentation for the `BP_REST_Components_Endpoint` I've noticed that the arguments of the generated documentation by the Restsplain plugin about the EDITABLE method are not consistent with what is actually expected.

It should be `name` and `action` instead of a subset of the item schema. See:
https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api-reference/endpoints/buddypress/v1/components/#PUT

NB: to ease WPCS testing I've also edited the composer.json file in the first commit, you can skip this commit if you think it's a bad idea ;)

NB 2 : The unit tests should error on Trunk, that's because of this:
https://buddypress.trac.wordpress.org/ticket/8115

@renatonascalves could you do a review of all the other endpoints to make sure arguments used in EDITABLE/CREATABLE/DELETABLE methods are consistent ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,5 +2,12 @@
     "name": "buddypress/bp-rest",
     "description": "WP REST API endpoints for BuddyPress",
     "type": "library",
-    "require": {}
+    "require": {
+        "squizlabs/php_codesniffer" : "^3.3.1"
+    },
+    "require-dev": {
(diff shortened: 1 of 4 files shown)


imath:
Just committed the patch on https://buddypress.trac.wordpress.org/ticket/8115 and re-ran the unit tests. There are now passing successfully.

PR #198, opened 2019-07-26
Make sure the collection params are consistent with BP_User_Query

imath:
For the members endpoint, we do not need all the `WP_REST_Users_Controller::get_collection_params()`. We only need the context, page, per_page and search. (who, roles, offset etc.. are not used by `BP_User_Query`).


imath:
FYI, in 686db10 I've added the arguments needed to create a member, the `email` was missing: it's required to create a user.

PR #201, opened 2019-07-31
Improve arguments description for the Activity endpoint methods

imath:
1. Make sure the `id` is listed as an argument for the `buddypress/v1/activity/<id>` and `buddypress/v1/activity/<id>/favorite` routes
2. Set the type of the content argument to `string` for CREATABLE and EDITABLE transport methods.
3. Set the type argument as required for the EDITABLE method (adapt the unit tests accordingly).
4. Improve strings used to describe the argument replacing "objetc" by "activity".
5. Set the status, comments, comment_count arguments as read only to remove them from EDITABLE and CREATABLE transport methods.
6. Set the title as readonly: activity actions are generated dynamically and remove the readonly property from the link argument: link is editable.

Note about the status: we will probably need specific routes to an spam/ham activity in a future release.

The [documentation for the activity endpoint](https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api/reference/activity/) has been written taking in account the above 6 points.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -66,6 +66,12 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			$this->namespace,
 			$activity_endpoint,
 			array(
+				'args'   => array(
+					'id' => array(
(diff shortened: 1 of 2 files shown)


renatonascalves:
will take a look later today! :)

PR #203, opened 2019-08-02
Improve Profile field arguments and some xProfile field features

imath:
1. Remove the duplicate `hide_empty_groups` parameter from the Group and Profile field endpoints.
2. Remove the `create_update_item_params()` method in favor of `get_endpoint_args_for_item_schema()` for consistency with other endpoints.
3. Make sure the ID is listed in arguments for the `/buddypress/v1/xprofile/fields/<id>` routes.
4. Set the default property of the `user_id` to `0` to avoid requesting for profile data by default for the GET transport method of `/buddypress/v1/xprofile/fields/<id>` route.
5. Introduce `get_endpoint_args_for_item_schema()` to build the schema for the EDITABLE and CREATABLE transport methods.
5.1 Make sure the visibility and autoling property of fields can be set using the endpoint.
5.2 Make sure `group_id`, `type`, and `name` are described as Required for the CREATABLE method.
5.3 Add a filter to eventually remove the specific arguments for visibility and autolink fields property.
6. Introduce a new method to set the visibility and autolink properties for the `create_item()` and `update_item()` methods.
7. Use `group_id` instead of `field_group_id` in REST requests (Update the unit tests accordingly).
8. Make sure it is possible to update fields property independently using the current value of the profile field object as fallback. This avoids resetting all properties when we only edit the description for example.
9. Make sure it is not possible to edit the `can_delete` property of the field ID: `bp_xprofile_fullname_field_id()`
10. Improve the descriptions of the item schema and make sure `readonly` and `default` property are set the right way.

NB: the [documentation for the Profile field endpoint](https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api/reference/extended-profiles/profile-field/) has been written taking in account the above 10 points.

### DIFF
--- a/includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
+++ b/includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
@@ -775,7 +775,7 @@ class BP_REST_XProfile_Field_Groups_Endpoint extends WP_REST_Controller {
 		);
 
 		$params['hide_empty_groups'] = array(
-			'description'       => __( 'True to hide groups that do not have any fields.', 'buddypress' ),
+			'description'       => __( 'Whether to hide profile groups of fields that do not have any profile fields or not.', 'buddypress' ),
 			'default'           => false,
 			'type'              => 'boolean',
 			'sanitize_callback' => 'rest_sanitize_boolean',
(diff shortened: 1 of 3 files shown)


includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
```diff
@@ -799,16 +799,8 @@ public function get_collection_params() {
 			'validate_callback' => 'bp_rest_validate_member_types',
 		);
 
-		$params['hide_empty_groups'] = array(
```
renatonascalves: good catch!

PR #204, opened 2019-08-03
Improve Profile data arguments and the way are retrieved field values

imath:
To preserve some consistency about data types and to make sure values are fully usable I have first edited the way field values are populating the REST responses (possibly into the Member, Profile Group, Profile Field and Profile Data endpoints):
- the `value` property is now an object exposing 3 versions of the value:
  - `raw` is the db version of the value (edit context only),
  - `unserialized` is always an array containing one or more field values (for instance checkboxes or multi-selectbox),
  - `rendered` is the HTML string output on the front-end. It may contain a link if the field has an `autolink` status set to `on`.
- an array is expected when saving a profile data. This is to take in account checkboxes or multi-selectbox which accept multiple values. If a string is used, it is transformed into an array before doing the inverse operation if the field type `! supports_multiple_defaults`.

To make this possible I had to introduce `BP_REST_XProfile_Fields_Endpoint->get_profile_field_unserialized_value()` and `BP_REST_XProfile_Fields_Endpoint->get_profile_field_rendered_value()` methods.

I have edited the corresponding item schemas and unit tests.

Back to Profile Data:
- I think it is important to have a `get_item()` method, otherwise it seems very weird not to be able to get Profile Data into the corresponding endpoint.
- I have switched from the `create_item()` method to the `update_item()` method (we have more verbs this way!) and most importantly it seems more consistent as the value is created as null during the Profile Field creation in a way.

Just like for the other endpoints, I have improved the arguments description and written the [endpoint documentation](https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api/reference/extended-profiles/profile-data/) accordingly to the above points.

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -495,7 +495,7 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 
 		// Get XProfile groups, only if the component is active.
 		if ( bp_is_active( 'xprofile' ) ) {
-			$field_enpoint = new BP_REST_XProfile_Fields_Endpoint();
+			$fields_endpoint = new BP_REST_XProfile_Fields_Endpoint();
 
 			$groups = bp_xprofile_get_groups(
 				array(
@@ -515,8 +515,8 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 						'name'  => $item->name,
 						'value' => array(
 							'raw'          => $item->data->value,
-							'unserialized' => $field_enpoint->get_profile_field_unserialized_value( $item->data->value ),
-							'rendered'     => $field_enpoint->get_profile_field_rendered_value( $item->data->value, $item ),
+							'unserialized' => $fields_endpoint->get_profile_field_unserialized_value( $item->data->value ),
+							'rendered'     => $fields_endpoint->get_profile_field_rendered_value( $item->data->value, $item ),
 						),
 					);
 				}
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -495,6 +495,8 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 
 		// Get XProfile groups, only if the component is active.
 		if ( bp_is_active( 'xprofile' ) ) {
+			$field_enpoint = new BP_REST_XProfile_Fields_Endpoint();
+
 			$groups = bp_xprofile_get_groups(
 				array(
 					'user_id'          => $user_id,
@@ -511,7 +513,11 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 				foreach ( $group->fields as $item ) {
 					$data['groups'][ $group->id ]['fields'][ $item->id ] = array(
 						'name'  => $item->name,
-						'value' => maybe_unserialize( $item->data->value ),
+						'value' => array(
+							'raw'          => $item->data->value,
+							'unserialized' => $field_enpoint->get_profile_field_unserialized_value( $item->data->value ),
+							'rendered'     => $field_enpoint->get_profile_field_rendered_value( $item->data->value, $item ),
+						),
 					);
 				}
 			}
--- a/includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
+++ b/includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
@@ -139,7 +139,7 @@ class BP_REST_XProfile_Data_Endpoint extends WP_REST_Controller {
 		$field = $this->get_xprofile_field_object( $request['field_id'] );
 
 		if ( empty( $field->id ) ) {
-			return new WP_Error(
+			$retval = new WP_Error(
 				'bp_rest_invalid_id',
 				__( 'Invalid field id.', 'buddypress' ),
 				array(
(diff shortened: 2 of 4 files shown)


includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -511,7 +513,11 @@ protected function xprofile_data( $user_id ) {
 				foreach ( $group->fields as $item ) {
 					$data['groups'][ $group->id ]['fields'][ $item->id ] = array(
 						'name'  => $item->name,
-						'value' => maybe_unserialize( $item->data->value ),
+						'value' => array(
+							'raw'          => $item->data->value,
+							'unserialized' => $field_enpoint->get_profile_field_unserialized_value( $item->data->value ),
```
renatonascalves:
```suggestion
							'unserialized' => $field_endpoint->get_profile_field_unserialized_value( $item->data->value ),
```

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -511,7 +513,11 @@ protected function xprofile_data( $user_id ) {
...
+							'rendered'     => $field_enpoint->get_profile_field_rendered_value( $item->data->value, $item ),
```
renatonascalves:
```suggestion
							'rendered'     => $field_endpoint->get_profile_field_rendered_value( $item->data->value, $item ),
```

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
```diff
@@ -63,14 +91,96 @@ public function register_routes() {
...
+
+	/**
+	 * Check if a given request has access to get users's data.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function get_item_permissions_check( $request ) {
```
renatonascalves: Did you notice you are losing the option of filtering the `bp_rest_xprofile_data_get_item_permissions_check`? The errors are being returned before the filter is being run.

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
```diff
@@ -403,18 +524,44 @@ public function get_item_schema() {
...
 					'readonly'    => true,
 					'type'        => 'integer',
 				),
 				'value'        => array(
 					'context'     => array( 'view', 'edit' ),
 					'description' => __( 'The value of the field data.', 'buddypress' ),
-					'type'        => 'integer',
+					'type'        => 'object',
+					'arg_options' => array(
+						'sanitize_callback' => null, // Note: sanitization implemented in self::prepare_item_for_database().
```
renatonascalves: This class doesn't have a `prepare_item_for_database` method. So this note seems to be inacurate. :/

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -843,6 +851,73 @@ public function get_xprofile_field_object( $request ) {
...
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  string                    $value         The raw value of the field.
+	 * @param  integer|BP_XProfile_Field $profile_field The ID or the full object for the field.
+	 * @return string                                   The field value for the display context.
+	 */
+	public function get_profile_field_rendered_value( $value = '', $profile_field = null ) {
+		if ( ! $value ) {
+			return '';
```
renatonascalves: I'm not sure what's the best practice is on cases where there is empty value, but is that common sense to pass it to the filter/hook?

If I'm hooking into `bp_get_the_profile_field_value` and even if it is empty, I never will see that because the hook doesn't return it.

renatonascalves: I think I misunderstood this one!

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Good catch! Thanks a lot 😍

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: 👍

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
imath: Thanks, I'll fix it asap!

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
imath: 👌

imath:
@renatonascalves it should be better now :) Thanks a lot for your review 👍

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
```diff
@@ -63,14 +91,96 @@ public function register_routes() {
...
+				'bp_rest_invalid_id',
+				__( 'Invalid field id.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+		}
+
+		// Check the user can view this field value.
+		$hidden_user_fields = bp_xprofile_get_hidden_fields_for_user( $request['user_id'] );
```
renatonascalves: I would confirm if the user exists first and return an error as well.

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -843,6 +851,73 @@ public function get_xprofile_field_object( $request ) {
...
+	 * Retrieve the unserialized value of a profile field.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  string $value The raw value of the field.
+	 * @return array         The unserialized field value.
+	 */
+	public function get_profile_field_unserialized_value( $value = '' ) {
+		if ( ! $value ) {
+			return '';
```
renatonascalves: The PHPDOC says it returns an array. So this should return an empty array?

```suggestion
			return array();
```

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
imath: I agree, thanks a lot for your feedback 👌

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
imath: Absolutely 😅

imath:
No problem @renatonascalves I've just committed to apply your recommandations 👍

PR #206, opened 2019-08-05
Group membership requests and invitations update.

dcavins:
Here's an updated PR with changes to the group membership requests and invitations endpoints.

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
@@ -247,7 +247,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 			// Do nothing.
 		} else {
 			$retval = new WP_Error(
-				'bp_rest_group_invites_cannot_get',
+				'bp_rest_group_invites_cannot_get_items',
 				__( 'Sorry, you are not allowed to fetch group invitations with those arguments.', 'buddypress' ),
 				array(
 					'status' => 500,
@@ -331,7 +331,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 						&& ! in_array( $user_id, array( $invite->user_id, $invite->inviter_id ), true )
 						) {
 				$retval = new WP_Error(
-					'bp_rest_authorization_required',
+					'bp_rest_group_invites_cannot_get_item',
 					__( 'Sorry, you are not allowed to fetch an invitation.', 'buddypress' ),
 					array(
 						'status' => 500,
@@ -386,7 +386,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 
 		if ( ! $invite_id ) {
 			return new WP_Error(
-				'bp_rest_group_invite_cannot_create',
+				'bp_rest_group_invite_cannot_create_item',
 				__( 'Could not invite member to the group.', 'buddypress' ),
 				array(
 					'status' => 500,
@@ -463,7 +463,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 		// Only a site admin or the user herself can extend invites.
 		} else if ( $inviter_id_arg !== bp_loggedin_user_id() && ! bp_current_user_can( 'bp_moderate' ) ) {
 			$retval = new WP_Error(
-				'bp_rest_group_invite_cannot_create',
+				'bp_rest_group_invite_cannot_create_item',
 				__( 'Sorry, you are not allowed to create the invitation as requested.', 'buddypress' ),
 				array(
 					'status' => 500,
@@ -478,7 +478,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 				'inviter_id' => $inviter_id_arg
 			) ) ) {
 				$retval = new WP_Error(
-					'bp_rest_group_invite_cannot_create',
+					'bp_rest_group_invite_cannot_create_item',
 					__( 'Sorry, you are not allowed to create the invitation as requested.', 'buddypress' ),
 					array(
 						'status' => 500,
@@ -514,7 +514,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 		$accept = groups_accept_invite( $invite->user_id, $invite->item_id );
 		if ( ! $accept ) {
 			return new WP_Error(
-				'bp_rest_group_invite_cannot_update',
+				'bp_rest_group_invite_cannot_update_item',
 				__( 'Could not accept group invitation.', 'buddypress' ),
 				array(
 					'status' => 500,
@@ -575,7 +575,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 				// Nothing to do.
 			} else if ( $user_id !== $invite->user_id	) {
 				$retval = new WP_Error(
-					'bp_rest_group_invite_cannot_update',
+					'bp_rest_group_invite_cannot_update_item',
 					__( 'Sorry, you are not allowed to accept the invitation as requested.', 'buddypress' ),
 					array(
 						'status' => 500,
@@ -618,7 +618,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 		$deleted   = groups_delete_invite( $invite->user_id, $invite->item_id, $invite->inviter_id );
 		if ( ! $deleted ) {
 			return new WP_Error(
-				'bp_rest_group_invite_cannot_delete',
+				'bp_rest_group_invite_cannot_delete_item',
 				__( 'Could not delete group invitation.', 'buddypress' ),
 				array(
 					'status' => 500,
@@ -682,7 +682,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 					&& ! groups_is_user_admin( $user_id, $invite->item_id )
 				) {
 				$retval = new WP_Error(
-					'bp_rest_group_invite_cannot_delete',
+					'bp_rest_group_invite_cannot_delete_item',
 					__( 'Sorry, you are not allowed to delete the invitation as requested.', 'buddypress' ),
 					array(
 						'status' => 500,
--- a/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
@@ -726,7 +726,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 			'invite_sent'   => $invite->invite_sent,
 			'inviter_id'    => $invite->inviter_id,
 			'group_id'      => $invite->item_id,
-			'date_modified' => $invite->date_modified,
+			'date_modified' => bp_rest_prepare_date_response( $invite->date_modified ),
 			'type'          => $invite->type,
 			'message'       => array(
 				'raw'      => $invite->content,
--- a/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
@@ -275,9 +275,8 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function get_item( $request ) {
-		$invites   = groups_get_invites( array( 'id' => $request['invite_id'] ) );
-		$invite    = current( $invites );
-		$retval    = $this->prepare_response_for_collection(
+		$invite = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$retval = $this->prepare_response_for_collection(
 			$this->prepare_item_for_response( $invite, $request )
 		);
 
@@ -306,10 +305,9 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 * @return bool|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-		$user_id   = bp_loggedin_user_id();
-		$invites   = groups_get_invites( array( 'id' => $request['invite_id'] ) );
-		if ( $invites ) {
-			$invite = current( $invites );
+		$user_id = bp_loggedin_user_id();
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		if ( $invite ) {
 			/**
 			 * Users can see a specific invitation if they
 			 * - are a site admin
@@ -508,10 +506,8 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 */
 	public function update_item( $request ) {
 		$user_id = bp_loggedin_user_id();
-		$invites = groups_get_invites( array( 'id' => $request['invite_id'] ) );
-		$invite  = current( $invites );
-
-		$accept = groups_accept_invite( $invite->user_id, $invite->item_id );
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$accept  = groups_accept_invite( $invite->user_id, $invite->item_id );
 		if ( ! $accept ) {
 			return new WP_Error(
 				'bp_rest_group_invite_cannot_update_item',
@@ -559,7 +555,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	public function update_item_permissions_check( $request ) {
 		$retval  = true;
 		$user_id = bp_loggedin_user_id();
-		$invites = groups_get_invites( array( 'id' => $request['invite_id'] ) );
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
 		if ( ! $user_id ) {
 			$retval = new WP_Error(
 				'bp_rest_authorization_required',
@@ -568,8 +564,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 					'status' => rest_authorization_required_code(),
 				)
 			);
-		} else if ( $invites ) {
-			$invite = current( $invites );
+		} else if ( $invite ) {
 			// Only the invitee or a site admin should be able to accept an invitation, and one must exist.
 			if ( bp_current_user_can( 'bp_moderate' ) ) {
 				// Nothing to do.
@@ -612,10 +607,9 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function delete_item( $request ) {
-		$user_id   = bp_loggedin_user_id();
-		$invites   = groups_get_invites( array( 'id' => $request['invite_id'] ) );
-		$invite    = current( $invites );
-		$deleted   = groups_delete_invite( $invite->user_id, $invite->item_id, $invite->inviter_id );
+		$user_id = bp_loggedin_user_id();
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$deleted = groups_delete_invite( $invite->user_id, $invite->item_id, $invite->inviter_id );
 		if ( ! $deleted ) {
 			return new WP_Error(
 				'bp_rest_group_invite_cannot_delete_item',
@@ -664,7 +658,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	public function delete_item_permissions_check( $request ) {
 		$retval  = true;
 		$user_id = bp_loggedin_user_id();
-		$invites = groups_get_invites( array( 'id' => $request['invite_id'] ) );
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
 		if ( ! $user_id ) {
 			$retval = new WP_Error(
 				'bp_rest_authorization_required',
@@ -673,8 +667,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 					'status' => rest_authorization_required_code(),
 				)
 			);
-		} else if ( $invites ) {
-			$invite = current( $invites );
+		} else if ( $invite ) {
 			// The inviter, the invitee, group admins, and site admins can all delete invites.
 			if ( bp_current_user_can( 'bp_moderate' ) ) {
 				// Nothing to do.
@@ -942,5 +935,21 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 		 * @param array $params Query params.
 		 */
 		return apply_filters( 'bp_rest_group_invites_collection_params', $params );
+	}
+
+	/**
+	 * Helper function to fetch a single group invite.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return BP_Invitation|bool $invite Invitation if found, false otherwise.
+	 */
+	public function fetch_single_invite( $args = array() ) {
+		$invites = groups_get_invites( $args );
+		if ( $invites ) {
+			return current( $invites );
+		} else {
+			return false;
+		}
 	}
 }
--- a/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
@@ -281,7 +281,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function get_item( $request ) {
-		$invite = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$invite = $this->fetch_single_invite( $request['invite_id'] );
 		$retval = $this->prepare_response_for_collection(
 			$this->prepare_item_for_response( $invite, $request )
 		);
@@ -312,7 +312,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 */
 	public function get_item_permissions_check( $request ) {
 		$user_id = bp_loggedin_user_id();
-		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$invite  = $this->fetch_single_invite( $request['invite_id'] );
 		$retval  = true;
 
 		if ( ! $user_id ) {
@@ -506,7 +506,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	 */
 	public function update_item( $request ) {
 		$user_id = bp_loggedin_user_id();
-		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$invite  = $this->fetch_single_invite( $request['invite_id'] );
 		$accept  = groups_accept_invite( $invite->user_id, $invite->item_id );
 		if ( ! $accept ) {
 			return new WP_Error(
@@ -555,7 +555,7 @@ class BP_REST_Group_Invites_Endpoint extends WP_REST_Controller {
 	public function update_item_permissions_check( $request ) {
 		$retval  = true;
 		$user_id = bp_loggedin_user_id();
-		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
+		$invite  = $this->fetch_single_invite( $request['invite_id'] );
 
 		if ( ! $user_id ) {
 			$retval = new WP_Error(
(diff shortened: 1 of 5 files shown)


includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -186,61 +260,145 @@ public function get_items_permissions_check( $request ) {
...
+	/**
+	 * Invite a member to a group.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param  WP_REST_Request $request Full data about the request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function create_item( $request ) {
+		$inviter_id_arg = $request['inviter_id'] ? $request['inviter_id'] : bp_loggedin_user_id();
```
renatonascalves: This fallback should be added via the params/collections as a default value.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -271,14 +430,69 @@ public function create_item( $request ) {
 	 * @return bool|WP_Error
 	 */
 	public function create_item_permissions_check( $request ) {
-		$retval = $this->get_items_permissions_check( $request );
+		$inviter_id_arg = $request['inviter_id'] ? $request['inviter_id'] : bp_loggedin_user_id();
```
renatonascalves: This fallback should be added via the params/collections as a default value.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -271,14 +430,69 @@ public function create_item( $request ) {
...
+		if ( ! is_user_logged_in() ) {
+			$retval = new WP_Error(
+				'bp_rest_authorization_required',
+				__( 'Sorry, you need to be logged in to create an invitation.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		} else if ( empty( $group->id ) ) {
+			return new WP_Error(
```
renatonascalves: This error is returning before the filter is being run.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -271,14 +430,69 @@ public function create_item( $request ) {
...
+				'bp_rest_group_invalid_id',
+				__( 'Invalid group id.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+		} else if ( ( empty( $user->ID ) || empty( $inviter->ID ) || $user->ID === $inviter->ID ) ) {
+			return new WP_Error(
```
renatonascalves: This error is returning before the filter is being run.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -449,16 +715,23 @@ public function delete_item_permissions_check( $request ) {
...
-			'user_id'      => $invite->user_id,
-			'invite_sent'  => $invite->invite_sent,
-			'inviter_id'   => $invite->inviter_id,
-			'is_confirmed' => $invite->is_confirmed,
+			'id'            => $invite->id,
+			'user_id'       => $invite->user_id,
+			'invite_sent'   => $invite->invite_sent,
+			'inviter_id'    => $invite->inviter_id,
+			'group_id'      => $invite->item_id,
+			'date_modified' => $invite->date_modified,
```
renatonascalves: Should we use `bp_rest_prepare_date_response` here?

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -292,56 +282,47 @@ public function get_item( $request ) {
...
+						'status' => 500,
+					)
+				);
 			}
-
-			if ( true === $retval ) {
-				$retval = true;
-			}
+		} else {
+			return new WP_Error(
```
renatonascalves: This is returning earlier than the filter.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -356,39 +337,42 @@ public function get_item_permissions_check( $request ) {
...
-		$group_member->is_confirmed  = 0;
-
-		if ( ! $group_member->save() ) {
+		$user_id_arg  = $request['user_id'] ? $request['user_id'] : bp_loggedin_user_id();
+		$user         = bp_rest_get_user( $user_id_arg );
+		$group        = $this->groups_endpoint->get_group_object( $request['group_id'] );
+
+		// Avoid duplicate requests.
+		if ( groups_check_for_membership_request( $user->ID, $group->id ) ) {
 			return new WP_Error(
```
renatonascalves: This is being returned before the hook.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -356,39 +337,42 @@ public function get_item_permissions_check( $request ) {
...
-		$group_member->group_id      = $group_id;
-		$group_member->user_id       = $user->ID;
-		$group_member->inviter_id    = 0;
-		$group_member->is_admin      = 0;
-		$group_member->user_title    = '';
-		$group_member->date_modified = bp_core_current_time();
-		$group_member->is_confirmed  = 0;
-
-		if ( ! $group_member->save() ) {
+		$user_id_arg  = $request['user_id'] ? $request['user_id'] : bp_loggedin_user_id();
```
renatonascalves: Default values should be passed via the params/collections.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -685,19 +624,132 @@ public function delete_item( $request ) {
...
+	 * @return WP_REST_Response
+	 */
+	public function prepare_item_for_response( $invite, $request ) {
+		$data = array(
+			'id'            => $invite->id,
+			'user_id'       => $invite->user_id,
+			'invite_sent'   => $invite->invite_sent,
+			'inviter_id'    => $invite->inviter_id,
+			'group_id'      => $invite->item_id,
+			'date_modified' => $invite->date_modified,
```
renatonascalves: same question here. Should we use `bp_rest_prepare_date_response`?

dcavins:
Hi @renatonascalves- Thanks again for the review. I'm sorry you don't like the permissions check. I went away from the `true &&` checks because I found them confusing and hard to read, unlike an if/else string which is easy to follow. However, if you prefer the other method, I can rework the checks again, since the REST API is really your baby. :) Thanks again!

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
dcavins: Thanks! Yes, after your last review, lots of items I had in permissions check vs the get/update/delete functions had to be slid around, and the premission check passes a $retval, where the others just return early.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: Ha! Since I've never heard of that function, I'm going to say "If you're suggesting it, it must be the way to go."

renatonascalves:
@dcavins If you feel strongly, keep it! Maybe only I have trouble with it! I can bet others will like the if/else pattern.

dcavins:
Nope! These endpoints should match the code style of the rest of the REST API. :)

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: The fallback only applies to the `create()` method. Does the WP REST API allow for defaults in one verb only?

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: Same as above.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: 👍

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: 👍

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
dcavins: 👍

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
dcavins: 👍

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -148,35 +182,75 @@ public function get_items( $request ) {
...
+		// Site administrators can do anything. Otherwise, the user must manage the subject group or be the requester.
+		} else if ( bp_current_user_can( 'bp_moderate' )
+				|| $request['group_id'] && $this->can_see( $request['group_id'] )
+				|| $user_id_arg === $user_id
+				|| $request['inviter_id'] === $user_id
+			) {
+			// Do nothing.
+		} else {
+			$retval = new WP_Error(
+				'bp_rest_group_invites_cannot_get',
```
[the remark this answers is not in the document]
dcavins: Sure, I can add `item(s)` to the error IDs.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
renatonascalves: It does!

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
renatonascalves: It does!

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -271,14 +436,61 @@ public function create_item( $request ) {
...
+			$retval = new WP_Error(
+				'bp_rest_group_invalid_id',
+				__( 'Invalid group id.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+		}
+
+		if ( true === $retval && ( empty( $user->ID ) || empty( $inviter->ID ) || $user->ID === $inviter->ID ) ) {
```
renatonascalves: Why `$user->ID === $inviter->ID` is invalid?

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -373,51 +610,42 @@ public function update_item_permissions_check( $request ) {
...
-				__( 'Invalid member id.', 'buddypress' ),
-				array(
-					'status' => 404,
-				)
-			);
-		}
-
-		$deleted = groups_delete_invite( $user->ID, $group->id );
+		$user_id = bp_loggedin_user_id();
+		$invite  = $this->fetch_single_invite( array( 'id' => $request['invite_id'] ) );
```
renatonascalves: Let me be a pain in the ass and suggest this method be `$this->fetch_single_invite( $request['invite_id']  );`

The `array( 'id' => $request['invite_id'] )` could be abastracted inside the method and not being called everytime.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -373,51 +610,42 @@ public function update_item_permissions_check( $request ) {
...
+		$user     = bp_rest_get_user( $invite->user_id );
+		$group    = $this->groups_endpoint->get_group_object( $invite->item_id  );
 
 		/**
 		 * Fires after a group invite is deleted via the REST API.
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param BP_Groups_Member $deleted_member  Deleted group member.
+		 * @param WP_User          $user            Subject user.
```
renatonascalves:
```suggestion
		 * @param WP_User          $user            User the group invite is being deleted from.
```

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -589,19 +900,38 @@ public function get_collection_params() {
...
+			'description'       => __( 'Return only invitations extended by this user.', 'buddypress' ),
+			'required'          => false,
+			'default'           => 0,
+			'type'              => 'integer',
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+
+		$params['invite_sent'] = array(
+			'description'       => __( 'Limit result set to invites that have been sent, not sent, or include all.', 'buddypress' ),
```
renatonascalves: If they only have those three options, maybe introduce an `enum` here so that we can whitelist them.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
 		 */
 		$args = apply_filters( 'bp_rest_group_membership_request_get_items_query_args', $args, $request );
 
-		$group    = $this->groups_endpoint->get_group_object( $args['group_id'] );
-		$user_ids = BP_Groups_Member::get_all_membership_request_user_ids( $group->id );
+		$memreqs = groups_get_requests( $args );
```
renatonascalves:
```suggestion
		$group_requests = groups_get_requests( $args );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
...
 
 		$retval = array();
-		foreach ( $user_ids as $user_id ) {
-			$group_member = new BP_Groups_Member( $user_id, $group->id );
-
+		foreach ( $memreqs as $memreq ) {
```
renatonascalves:
```suggestion
		foreach ( $group_requests as $group_request ) {
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
...
 			$retval[] = $this->prepare_response_for_collection(
-				$this->membership_endpoint->prepare_item_for_response( $group_member, $request )
+				$this->prepare_item_for_response( $memreq, $request )
```
renatonascalves:
```suggestion
				$this->prepare_item_for_response( $group_request, $request )
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
...
 			);
 		}
 
 		$response = rest_ensure_response( $retval );
-		$response = bp_rest_response_add_total_headers( $response, count( $user_ids ), $args['per_page'] );
+		$response = bp_rest_response_add_total_headers( $response, count( $memreqs ), $args['per_page'] );
```
renatonascalves:
```suggestion
		$response = bp_rest_response_add_total_headers( $response, count( $group_requests ), $args['per_page'] );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
...
-		 * @param array            $user_ids     List of membership requests.
-		 * @param BP_Groups_Group  $group        The group object.
-		 * @param WP_REST_Response $response     The response data.
-		 * @param WP_REST_Request  $request      The request sent to the API.
+		 * @param array of BP_Invitations $memreqs      List of membership requests.
+		 * @param WP_REST_Response        $response     The response data.
+		 * @param WP_REST_Request         $request      The request sent to the API.
 		 */
-		do_action( 'bp_rest_group_membership_request_get_items', $user_ids, $group, $response, $request );
+		do_action( 'bp_rest_group_membership_request_get_items', $memreqs, $response, $request );
```
renatonascalves:
```suggestion
		do_action( 'bp_rest_group_membership_request_get_items', $group_requests, $response, $request );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -158,32 +144,28 @@ public function get_items( $request ) {
...
 		/**
 		 * Fires after a list of group membership request is fetched via the REST API.
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param array            $user_ids     List of membership requests.
-		 * @param BP_Groups_Group  $group        The group object.
-		 * @param WP_REST_Response $response     The response data.
-		 * @param WP_REST_Request  $request      The request sent to the API.
+		 * @param array of BP_Invitations $memreqs      List of membership requests.
```
renatonascalves:
```suggestion
		 * @param array of BP_Invitations $group_requests      List of membership requests.
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -685,19 +634,134 @@ public function delete_item( $request ) {
...
+
+		/**
+		 * Filter links prepared for the REST response.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param array    $links  The prepared links of the REST response.
+		 * @param stdClass $invite Invite object.
+		 */
+		return apply_filters( 'bp_rest_group_invites_prepare_links', $links, $invite );
```
renatonascalves:
```suggestion
		return apply_filters( 'bp_rest_group_membership_request_prepare_links', $links, $invite );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -685,19 +634,134 @@ public function delete_item( $request ) {
...
+		 */
+		return apply_filters( 'bp_rest_group_membership_requests_prepare_value', $response, $request, $invite );
+	}
+
+	/**
+	 * Prepare links for the request.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param stdClass $invite Invite object.
```
renatonascalves:
```suggestion
	 * @param BP_Invitation $invite Invite object.
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -685,19 +634,134 @@ public function delete_item( $request ) {
...
+			),
+		);
+
+		/**
+		 * Filter links prepared for the REST response.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param array    $links  The prepared links of the REST response.
+		 * @param stdClass $invite Invite object.
```
renatonascalves:
```suggestion
		 * @param BP_Invitation $invite Invite object.
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -685,19 +634,134 @@ public function delete_item( $request ) {
...
+		/**
+		 * Filter a group invite value returned from the API.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param WP_REST_Response $response The response data.
+		 * @param WP_REST_Request  $request  Request used to generate the response.
+		 * @param BP_Invitation    $invite   The invite object.
+		 */
+		return apply_filters( 'bp_rest_group_membership_requests_prepare_value', $response, $request, $invite );
```
renatonascalves:
```suggestion
		return apply_filters( 'bp_rest_group_membership_requests_prepare_value', $response, $request, $invite );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -720,4 +784,65 @@ public function get_item_schema() {
...
+			'sanitize_callback' => 'sanitize_text_field',
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+
+		/**
+		 * Filters the collection query params.
+		 *
+		 * @param array $params Query params.
+		 */
+		return apply_filters( 'bp_rest_group_invites_collection_params', $params );
```
renatonascalves:
```suggestion
		return apply_filters( 'bp_rest_group_membership_requests_collection_params', $params );
```

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -720,4 +784,65 @@ public function get_item_schema() {
...
+	}
+
+	/**
+	 * Helper function to fetch a single group invite.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @return BP_Invitation|bool $memreq Membership request if found, false otherwise.
+	 */
+	public function fetch_single_invite( $args = array() ) {
```
renatonascalves: Same as before!

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: A user cannot invite himself to a group. "user" is the invited user, "inviter" is the user extending the invitation.

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: Ha ha ha, what's "pain in the ass" in Portuguese? :)

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
dcavins: 👍

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
dcavins: I can't see what is changed here?

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
renatonascalves: Annoying or inconvenient person!

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
renatonascalves: My mistake!

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -293,39 +505,29 @@ public function create_item_permissions_check( $request ) {
...
-				array(
-					'status' => 404,
-				)
-			);
-		}
-
-		$accept = groups_accept_invite( $user->ID, $group->id );
+		$user_id = bp_loggedin_user_id();
+		$invite  = $this->fetch_single_invite( $request['invite_id'] );
+		$accept  = groups_accept_invite( $invite->user_id, $invite->item_id );
```
imath: So the `update_item()` method is to accept an invite, I'm wondering (sorry for this late wake up!) if the `update_item()` method should also include an argument like
```php
$args['action'] = array(
  'description' => __( 'Whether to accept or reject a Group invite', 'buddypress' ),
  'type'          => 'string',
  'enum'        => array( 'accept', 'reject' ),
  'default'      => 'accept',
}
```
and use this argument to eventually reject the invite. Because, otherwise I'm afraid `do_action( 'groups_reject_invite' )` will never be triggered into the BP REST API.

Or am I missing something?

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
```diff
@@ -373,51 +610,42 @@ public function update_item_permissions_check( $request ) {
...
-				array(
-					'status' => 404,
-				)
-			);
-		}
-
-		$deleted = groups_delete_invite( $user->ID, $group->id );
+		$user_id = bp_loggedin_user_id();
+		$invite  = $this->fetch_single_invite( $request['invite_id'] );
+		$deleted = groups_delete_invite( $invite->user_id, $invite->item_id, $invite->inviter_id );
```
imath: It looks like `groups_delete_invite()` is used to clean the DB, more than an action made by a user:
if an invited member to a public group clicks on the `Join Group` button instead of accepting the invite, then we use this function.

I think `update_item()` should take care of accept/reject (see previous comment) and `delete_item()` should handle `groups_uninvite_user()` : when an inviter changed his mind and wants to remove the invite he sent.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -356,39 +344,42 @@ public function get_item_permissions_check( $request ) {
...
 			return new WP_Error(
-				'bp_rest_group_membership_request_failed',
-				__( 'Could not send membership request to this group.', 'buddypress' ),
+				'bp_rest_group_membership_requests_duplicate_request',
+				__( 'There is already a request to this member.', 'buddypress' ),
 				array(
 					'status' => 500,
 				)
 			);
 		}
```
imath: What about joining a public group ? I think we are simply not handling this case 😝 
It's a kind of membership request already accepted. Maybe we should check the status of the group and use `groups_join_group()` for public group and eventually check for existing invites before using `groups_delete_invite()` if one was found ?

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
imath: Another thing, in Core we are also using `groups_delete_membership_request()` to clean existing membership request...

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -504,41 +482,40 @@ public function create_item_permissions_check( $request ) {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function update_item( $request ) {
-		$user  = bp_rest_get_user( $request['user_id'] );
-		$group = $this->groups_endpoint->get_group_object( $request['group_id'] );
-
-		if ( ! groups_accept_membership_request( false, $user->ID, $group->id ) ) {
+		$group_request = $this->fetch_single_invite( $request['request_id'] );
+		$success = groups_accept_membership_request( false, $group_request->user_id, $group_request->item_id );
+		if ( ! $success ) {
```
imath: Just like for invites, we should also handle `groups_reject_membership_request()` here I think

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
```diff
@@ -637,41 +585,42 @@ public function update_item_permissions_check( $request ) {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function delete_item( $request ) {
-		$user  = bp_rest_get_user( $request['user_id'] );
-		$group = $this->groups_endpoint->get_group_object( $request['group_id'] );
-
-		if ( ! groups_reject_membership_request( false, $user->ID, $group->id ) ) {
+		$group_request = $this->fetch_single_invite( $request['request_id'] );
+		$success       = groups_reject_membership_request( false, $group_request->user_id, $group_request->item_id );
```
imath: This should be to let a user who requested a membership to delete his request with `groups_delete_membership_request()` instead of reject.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
imath: Fixed in https://github.com/buddypress/BP-REST/commit/e8da8b20a9bc7c83bf185f6bf4f3a1f6c2b08094

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
imath: As seen during dev chat: it's a wrong approach. Responses would be too different. `groups_uninvite_user()` and `groups_reject_invite()` should probably be handled into the delete_item() method

includes/bp-groups/classes/class-bp-rest-group-invites-endpoint.php
imath: As seen during dev chat: it's a wrong approach.

includes/bp-groups/classes/class-bp-rest-group-membership-request-endpoint.php
imath: As seen during dev chat: it's a wrong approach. `groups_reject_membership_request()` should probably be handled in the `delete_method()`

renatonascalves:
Rerun Travis and tests are passing now. :)

imath:
Thanks ! I think we should rebase & merge and improve it if we need to later. Any objections if I do so in the coming hours?

renatonascalves:
@imath not at all! As soon as it is in, the better! :)

PR #210, opened 2019-08-08
Harmonize delete_item() method responses

imath:
As discussed during August 7th dev chat, I think it's important we harmonize how we build the responses of the endpoints `delete_item()` methods.

I've suggested to do just like the WP REST API does for this case. One good reason to do so is: our Members endpoint extends the WP Users one and already uses this specific way to build responses. See the extract of `WP_REST_Users_Controller->delete_item()` code below:

```php
		$result = wp_delete_user( $id, $reassign );

		if ( ! $result ) {
			return new WP_Error( 'rest_cannot_delete', __( 'The user cannot be deleted.' ), array( 'status' => 500 ) );
		}

		$response = new WP_REST_Response();
		$response->set_data(
			array(
				'deleted'  => true,
				'previous' => $previous->get_data(),
			)
		);
```

This PR is applying this to all endpoints except for the 2 ones @dcavins is currently working on in #206

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -607,11 +607,11 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function delete_item( $request ) {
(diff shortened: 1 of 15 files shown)


dcavins:
I think makes sense. Thanks for making the changes!

PR #212, opened 2019-08-10
Improve/notifications endpoints arguments

imath:
1. Adapt the item schema replacing "object" by notification and making sure needed arguments by the `CREATABLE` transport method can be used.
2. Make sure a logged in user can list his notifications.
3. Make sure it is possible to change the sort order of the Notifications query.
4. Make sur the ID argument is listed in the parameters of the route `buddypress/v1/notifications/<id>`
5. Only keep the `id` and `is_new` arguments as it is only possible to toggle the status of a notification.
6. Update the `is_new` property of the notification object before building the response.
7. Adapt unit tests according to introduced improvements.

The [Notifications Endpoints documentation](https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api/reference/screen-notifications/) has been written taking in account these changes.

### DIFF
--- a/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
+++ b/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
@@ -98,10 +98,9 @@ class BP_REST_Notifications_Endpoint extends WP_REST_Controller {
 			'secondary_item_id' => $request['secondary_item_id'],
 			'component_name'    => $request['component_name'],
 			'component_action'  => $request['component_action'],
-			'date_query'        => $request['date'],
-			'sort_order'        => $request['sortby'],
+			'order_by'          => $request['order_by'],
+			'sort_order'        => strtoupper( $request['sort_order'] ),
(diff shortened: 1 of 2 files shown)


includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
```diff
@@ -83,6 +89,26 @@ public function register_routes() {
...
+	public function get_endpoint_args_for_item_schema( $method = WP_REST_Server::CREATABLE ) {
+		$args = WP_REST_Controller::get_endpoint_args_for_item_schema( $method );
+
+		if ( WP_REST_Server::EDITABLE === $method ) {
+			// Only switching the is_new property can be achieved.
+			$args                      = array_intersect_key( $args, array( 'is_new' => true ) );
+			$args['is_new']['default'] = 0;
+		}
+
+		return $args;
```
renatonascalves: MIssing hook...

imath:
Thanks a lot for your review @renatonascalves I've just added the missing filter 👍

PR #221, opened 2019-08-15
Invitations: Take current user into account when deleting requests an…

dcavins:
…d invitations.

BuddyPress offers a variety of delete functions for invitations and requests, representing different situations.
• `groups_uninvite_user()`
• `groups_reject_invite()`
• `groups_delete_invite()`
• `groups_reject_membership_request()`
• `groups_delete_membership_request()`

I'm assuming that the situation is driven by who is doing the deleting. In this patch, we compare the current user to the subject user of the invitation or request to determine whether the invitation or request is being withdrawn or rejected.


imath:
It looks like #222 includes this commit, maybe de should close this PR?

dcavins:
> It looks like #222 includes this commit, maybe we should close this PR?

Whatever you think is best. While I was in there on the context thing, I thought I'd better fix the response while I was thinking about it. :D

PR #233, opened 2019-08-26
Improve the Group Avatar Controller

imath:
- Make sure the get_item response is consistent with the item schema. That is to say: always return the full and thumb version of the avatar.
- Add the `action` argument to the CREATABLE transport method.
- Add a unit test for the `create_item` method.

Write the [documentation](https://imath-buddydocs.pf1.wpserveur.net/bp-rest-api/reference/attachments/group-avatar/) about this controller taking in account these changes.


tests/attachments/test-group-avatar-controller.php
```diff
@@ -90,9 +97,68 @@ public function test_get_item_with_no_image() {
 
 	/**
 	 * @group create_item
+	 * @group imath
```
renatonascalves:
```suggestion
```

imath:
Ahah! Good catch, thanks a lot 👌

PR #240, opened 2019-09-01
Fix some i18n issues

imath:
1. There was 2 missing text domains (`buddypress`).
2. `id` or `ID`.. I chose `ID` everywhere :)
3. I also tried to limit ways of writing the same thing. Eg: not authorized ~ unauthorized or sometimes a full stop was missing.
4. I fixed a misuse of the %1$s placeholder
5. I fixed a translator comment where %$1s was used instead of %s


imath:
Thanks for your review. I just applyed requested change 👌

PR #241, opened 2019-09-24
added filter for activity data

satheshk385:
added filter to change activity data, because we can't filter activity response data

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -947,6 +947,8 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 				),
 			);
 		}
+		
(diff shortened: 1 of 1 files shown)


renatonascalves:
@satheshk385 Have you tried using this hook? https://github.com/buddypress/BP-REST/blob/28dda16881232769b4cef3b9fa2f901e1492c819/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php#L969

You can filter the information via the `bp_rest_activity_prepare_value` hook...

satheshk385:
@renatonascalves 
I have tried, But we can't add data in that $response variable.

checkout the class reference of $response object
https://developer.wordpress.org/reference/classes/wp_rest_response/
https://developer.wordpress.org/reference/classes/wp_rest_response/

satheshk385:
@renatonascalves 
you are right we can filter data
$response->get_data();
$response->set_data();
https://developer.wordpress.org/reference/classes/wp_http_response/

renatonascalves:
:)

PR #244, opened 2019-10-23
Use the right check for group avatars

imath:
The check for group avatars should be `! bp_disable_group_avatar_uploads()` and it should be checked when preparing the response.

**This needs to be merged into Master & 0.1**

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -597,31 +597,36 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 			'name'               => bp_get_group_name( $item ),
 			'slug'               => bp_get_group_slug( $item ),
 			'status'             => bp_get_group_status( $item ),
-			'avatar_urls'        => array(),
 			'admins'             => array(),
 			'mods'               => array(),
 			'total_member_count' => null,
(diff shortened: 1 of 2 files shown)


includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -1051,7 +1056,7 @@ public function get_item_schema() {
 		);
 
 		// Avatars.
-		if ( true === buddypress()->avatar->show_avatars ) {
+		if ( ! bp_disable_group_avatar_uploads() ) {
```
renatonascalves: I see! This is a good opportunity to change in other places as well, as I used this pattern, `if ( true === buddypress()->avatar->show_avatars ) {` in other places.

imath: Well, this can be done in another PR imho

imath:
The failing test is due to a failing svn co in the server. It should be ok it you restart the job

imath:
I've restarted the job and the test is fine. What was the minor change @renatonascalves ?

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
imath: `true === buddypress()->avatar->show_avatars` is the right check when fetching member avatars

PR #247, opened 2019-10-25
Improve Avatar & Cover image controllers conditional loading

imath:
1. The User Avatar is attached to the xProfile component (not the Members one)
2. Only load Avatar/Cover image uploads if local uploads are enabled for both objects
3. The cover image needs an additional check because some theme might deactivate it removing their support to it (it is the purpose of bp_is_active( group, cover_image )).

PS: as this part is handled within Components class into BuddyPress Core, there is no need to backport the commit on branch 0.1 imho.

### DIFF
--- a/bp-rest.php
+++ b/bp-rest.php
@@ -58,11 +58,6 @@ function bp_rest() {
 		$controller = new BP_REST_Members_Endpoint();
 		$controller->register_routes();
 
-		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/trait-attachments.php';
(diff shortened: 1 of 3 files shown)


imath:
FYI for BuddyPress 5.1.0 we'll use https://buddypress.trac.wordpress.org/ticket/8151

imath:
Important note: changes about:
- includes/bp-attachments/classes/class-bp-rest-attachments-member-avatar-endpoint.php
- tests/attachments/test-member-avatar-controller.php

will need to be back ported to 0.1

imath:
Looks like it's fine to do things differently from what it is in BuddyPress Core.. So let's forget this PR. For more details; you can read this discussion:
https://wordpress.slack.com/archives/C02RQBYUG/p1572193392178200

PR #250, opened 2019-10-31
Load the Member Avatar Controller if the xProfile component is active

### LINKED ISSUE #249: Uploading or deleting an avatar is not possible when the xProfile component is not active
Following up the discussion we had on Slack about the fact it "makes more sense" to attach the local avatar upload to the Members component in the BP REST API although it's attached to the xProfile component in BuddyPress Core, I've just tested if it was possible to upload/delete a local avatar when the xProfile component was not active. And it's actually not possible. As my way to resolve this (attaching the feature just like it is in BuddyPress Core to the xProfile component) doesn't make sense for the BP REST API. Thanks for working on resolving this issue.

Here are the steps to reproduce :

## Uploading

1. Log in as the site administrator.
2. Deactivate the xProfile component.
3. Use this [example of use](https://developer.buddypress.org/bp-rest-api/reference/attachments/member-avatar/#example-of-use-2) editing the user ID into the REST URL according to the one of the logged in user .
4. Go to any post and try to upload a new image.
5. View the following result into the browser console:

xprofile-dependency

## Deleting

6. Change the code in 3. and use this [one instead](https://developer.buddypress.org/bp-rest-api/reference/attachments/member-avatar/#example-of-use-3)
7. Refresh the page
8. View the following result into the browser console:

xprofile-dependency-delete

If you re-activate the xProfile component and redo the steps above, you'll see it's working the right way.

imath:
As discussed during October 30 dev-chat: until the Member Avatar dependency to xProfile is removed in BuddyPress Core, we will make sure to load the Member Avatar Controller only if the xProfile component is active.

Documentation about this fact has been updated, see [Member Avatar chapter](https://developer.buddypress.org/bp-rest-api/reference/attachments/member-avatar/).

Fixes #249

### DIFF
--- a/bp-rest.php
+++ b/bp-rest.php
@@ -58,11 +58,6 @@ function bp_rest() {
 		$controller = new BP_REST_Members_Endpoint();
 		$controller->register_routes();
 
-		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/trait-attachments.php';
-		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/class-bp-rest-attachments-member-avatar-endpoint.php';
-		$controller = new BP_REST_Attachments_Member_Avatar_Endpoint();
-		$controller->register_routes();
-
 		if ( bp_get_signup_allowed() ) {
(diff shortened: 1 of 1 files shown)


renatonascalves:
Awesome! :)

PR #253, opened 2019-11-08
Blogs Avatar Endpoint

### LINKED ISSUE #245: The Blogs Controller should fetch Blog avatars
We'll need to wait for https://buddypress.trac.wordpress.org/ticket/8150 to be fixed. To fetch the blog avatar it will be required to use the `bp_get_blog_avatar()` because it's checking for the site icon to eventually replace the avatar of the Blog's admin by it.

renatonascalves:
fixes #245 

This PR has the blog avatar endpoint and also the `get_item` endpoint for the Blog component. Meaning one can now fetch a single blog data via the API.

Also I'm adding the avatar information to the response if allowed.


renatonascalves:
Unit tests are not working because https://buddypress.trac.wordpress.org/ticket/8150 patch wasn't merged in yet.

bp-rest.php
```diff
@@ -75,6 +75,14 @@ function bp_rest() {
 		require_once dirname( __FILE__ ) . '/includes/bp-blogs/classes/class-bp-rest-blogs-endpoint.php';
 		$controller = new BP_REST_Blogs_Endpoint();
 		$controller->register_routes();
+
+		// Support to Blog Avatar.
+		if ( bp_is_active( 'blogs', 'site-icon' ) ) {
```
imath: I wonder if it would be better to use something like `bp_is_active( 'blogs', 'avatar' )` instead and add support for it in BuddyPress trunk... The site icon is more a fallback before the admin user's avatar fallback imho.

includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
```diff
@@ -857,7 +857,7 @@ public function get_collection_params() {
 		);
 
 		$params['exclude_groups'] = array(
-			'description'       => __( 'Ensure result set excludes specific profile field groups.', 'buddypress' ),
+			'description'       => __( '', 'buddypress' ),
```
imath: You should probably remove this file from the PR as it deals with xProfiles and not Blogs :)

tests/attachments/test-group-avatar-controller.php
```diff
@@ -51,7 +51,7 @@ public function test_register_routes() {
 	 * @group get_items
```
imath: The changes in this file should be in another PR imho.

tests/attachments/test-group-cover-controller.php
```diff
@@ -44,14 +44,14 @@ public function test_register_routes() {
 	 * @group get_items
```
imath: The changes in this file should be in another PR imho.

tests/attachments/test-member-avatar-controller.php
```diff
@@ -45,7 +45,7 @@ public function test_register_routes() {
 	 * @group get_items
```
imath: The changes in this file should be in another PR imho.

bp-rest.php
renatonascalves: @imath what about back-compact?

imath: You're right, forget about this comment :) It's fine the way it is.

imath:
@renatonascalves I believe we've lost enough time with using `filesize()` when WP < 4.9, let's move forward skipping the tests. Also I believe testing 4.7 is no more needed as BP supports WP >= 4.8.

renatonascalves:
@imath So you pinpointed the problem to the `filesize` function on WP < 4.9. Is that it?

imath:
@renatonascalves I've tried and finally the last commit is skipping the problematic tests if WordPress is below 4.9

PR #255, opened 2019-11-16
Showing hidden groups correctly

renatonascalves:
This make sure that only admins and hidden group members can see the hidden group(s).

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -135,6 +135,12 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 
 		$retval = array();
 		foreach ( $groups['groups'] as $group ) {
+
+			// Remove hidden groups if the user can't see it.
+			if ( true === $request['show_hidden'] && ! $this->can_see_hidden_groups( $group ) ) {
+				continue;
+			}
+
 			$retval[] = $this->prepare_response_for_collection(
 				$this->prepare_item_for_response( $group, $request )
 			);
@@ -166,17 +172,6 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 	 * @return bool|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
-		$retval = true;
-
-		if ( ! $this->can_see_hidden_groups( $request ) ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you cannot view hidden groups.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		}
 
 		/**
 		 * Filter the groups `get_items` permissions check.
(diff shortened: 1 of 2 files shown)


tests/groups/test-controller.php
```diff
@@ -102,6 +102,46 @@ public function test_get_paginated_items() {
...
+		);
+
+		$request->set_param( 'context', 'view' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 200, $response->get_status() );
+
+		$all_data = $response->get_data();
+
+		$this->assertTrue( count( $all_data ) === 2 );
```
renatonascalves: This is correct because we add a global group on `setUp()`... So one public group plus the one hidden from this test, makes it two.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -135,6 +135,12 @@ public function get_items( $request ) {
 
 		$retval = array();
 		foreach ( $groups['groups'] as $group ) {
+
+			// Remove hidden groups if the user can't see it.
+			if ( true === $request['show_hidden'] && ! $this->can_see_hidden_groups( $group ) ) {
+				continue;
+			}
+
```
imath: The disadvantage of this is that it's going to cause an inconsistent count. After checking it more deeply, we're actually only including hidden groups if the logged in user is the super admin on the "All groups" context and we include hidden groups for every logged in users when on the "My Groups" context. See the following code on [SVN Trunk](https://buddypress.trac.wordpress.org/browser/trunk/src/bp-groups/classes/class-bp-groups-template.php#L186)

```php
186	                if ( bp_current_user_can( 'bp_moderate' ) || ( is_user_logged_in() && $user_id == bp_loggedin_user_id() ) ) {
187	                        $show_hidden = true;
188	                }
```

So I think we should probably do the same here. Meaning:

1. if $request['user_id'] is set and equals the logged in user, then set show_hidden to true.
2. if logged in user can `bp_moderate` then set the show_hidden to true.
3. if show_hidden is true and $request['user_id'] is not the logged in user id and cannot `bp_moderate` then reset show_hidden to false.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -166,17 +172,6 @@ public function get_items( $request ) {
...
-
-		if ( ! $this->can_see_hidden_groups( $request ) ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you cannot view hidden groups.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		}
```
imath: I agree this should be removed 👍

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -841,24 +836,25 @@ protected function can_user_delete_or_update( $group ) {
...
 	 * @since 0.1.0
 	 *
-	 * @param WP_REST_Request $request Full details about the request.
+	 * @param BP_Groups_Groupt $group Group object.
+	 *
 	 * @return bool
 	 */
-	protected function can_see_hidden_groups( $request ) {
-		if ( $request['show_hidden'] ) {
+	protected function can_see_hidden_groups( $group ) {
```
imath: I believe this function should only be used for the `get_item()` method.

tests/groups/test-controller.php
imath: Maybe add a comment above ? or something like this:

```php
$this->assertTrue( count( $all_data ) === 2, 'Total should be 2 as a group is added on setUp()' );
```

renatonascalves:
Tks! In that case, the current implementation was right all along. :)

PR #257, opened 2019-12-29
Fix 2 issues about the `edit` context for the Groups controller get_items() method

imath:
1) Regular users should not be able to use the `edit` context.
2) Some user data should be unset like the user_pass or the user_email when building the list of mods/admins/members inside the `edit` context.

**Important**:  we'll need to backport the fix on [branch 0.1](https://github.com/buddypress/BP-REST/tree/0.1)

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -178,6 +178,11 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 			);
 		}
 
+		// Users need (at least, should we be more restrictive ?) to be logged in to use the edit context.
+		if ( 'edit' === $request->get_param( 'context' ) && ! is_user_logged_in() ) {
+			$request->set_param( 'context', 'view' );
+		}
+
 		/**
 		 * Filter the groups `get_items` permissions check.
 		 *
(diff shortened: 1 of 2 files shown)


tests/groups/test-controller.php
```diff
@@ -102,6 +102,59 @@ public function test_get_paginated_items() {
...
+
+		$admins = array();
+		$groups = $response->get_data();
+		foreach ( $groups as $group ) {
+			if ( isset( $group['admins'] ) ) {
+				$admins = array_merge( $admins, $group['admins'] );
+			}
+		}
+
+		$this->assertEmpty( $admins, 'Listing Admins should not be possible for regular users' );
```
renatonascalves: The regular user part bugs me. I'd rather put something making clear that non-members can't see it with the `edit` context. Regular users seem to be normal users, which might confuse people.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -647,6 +652,20 @@ public function prepare_item_for_response( $item, $request ) {
 			);
 
 			foreach ( (array) $admin_mods['members'] as $user ) {
+				// Make sure to unset private data.
+				$private_keys = array_intersect(
```
renatonascalves: Good catch! Didn't test this thoughly at the time. Actually, I assumed those information were not there.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -178,6 +178,11 @@ public function get_items_permissions_check( $request ) {
 			);
 		}
 
+		// Users need (at least, should we be more restrictive ?) to be logged in to use the edit context.
+		if ( 'edit' === $request->get_param( 'context' ) && ! is_user_logged_in() ) {
```
renatonascalves: I'm okay adding this in the method perfoming the action instead of the permissions check.

tests/groups/test-controller.php
imath: Fine with me :)

imath:
Hi @renatonascalves thanks for your feedback. I've just applied your recommandations 👍

imath:
FYI, I've just backported the fix on the 0.1 branch. See c75e0a5187e7f289955ae018d7ea886852f94c26.

renatonascalves:
Sounds great! Thank you! :)

PR #277, opened 2020-03-01
Update Attachments implementation

renatonascalves:
Updating attachments implementation to account for the move from the XProfile to the Members component.

### DIFF
--- a/bp-rest.php
+++ b/bp-rest.php
@@ -58,6 +58,18 @@ function bp_rest() {
 		$controller = new BP_REST_Members_Endpoint();
 		$controller->register_routes();
 
+		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/trait-attachments.php';
+
+		// Support Member's Avatar.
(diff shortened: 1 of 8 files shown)


bp-rest.php
```diff
@@ -58,6 +58,18 @@ function bp_rest() {
...
 		$controller->register_routes();
 
+		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/trait-attachments.php';
+
+		// Support Member's Avatar.
+		require_once dirname( __FILE__ ) . '/includes/bp-attachments/classes/class-bp-rest-attachments-member-avatar-endpoint.php';
+		$controller = new BP_REST_Attachments_Member_Avatar_Endpoint();
+		$controller->register_routes();
+
+		// Support Member's Cover.
```
imath: There’s should be a check for:
```php
bp_is_active( 'members', 'cover_image' )
```

renatonascalves: Good catch! Added!

renatonascalves:
@imath Yes! We are confirming if upload is disabled for avatar and cover images. 

That reminded me of adding unit tests for those functionalities and they were added now. :)

PR #283, opened 2020-03-07
Friendship Improvements

### LINKED ISSUE #280: Cannot remove friend via friends endpoint
The friends endpoint allows creating friendship requests, accepting friendship requests, rejecting friendship requests, withdrawing friendship requests but it does not allow removing friends.

Performing a DELETE on `/friends/<friendship-id>` endpoint deals with friendship requests, not friendships themselves. Either `friends_withdraw_friendship()` or `friends_reject_friendship()` is called, `friends_remove_friend()` is never called within the plugin.

`friends_withdraw_friendship()` will not delete confirmed friendships.

renatonascalves:
fixes #280

### DIFF
--- a/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
+++ b/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
@@ -458,7 +458,7 @@ class BP_REST_Friends_Endpoint extends WP_REST_Controller {
 	}
 
 	/**
-	 * Reject/withdraw friendship.
+	 * Reject/withdraw/remove friendship.
 	 *
 	 * @since 6.0.0
 	 *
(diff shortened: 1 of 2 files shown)


sjregan:
@renatonascalves Thanks for your work on this, however I am not sure it fixes the problem.

`get_endpoint_args_for_item_schema()` isn't called with `WP_REST_Server::DELETABLE` as a parameter, which means `rest_sanitize_boolean` isn't applied to the `force` parameter so it never equals true.

renatonascalves:
@sjregan Did you test this PR locally or otherwise? I'm able to delete with the `force` flag which this PR is introducing.

sjregan:
> @sjregan Did you test this PR locally or otherwise? I'm able to delete with the `force` flag which this PR is introducing.

@renatonascalves I'm tested locally, making requests with Postman. Which means the WP_REST_Request parameters aren't being set via code as they are in the test suites, so the 'force' param remains a string, rather than being converted to a boolean via `rest_sanitize_boolean` as one would expect.

Looking at the code I can see `get_endpoint_args_for_item_schema` is not called when registering the delete_item endpoint (it is called for the create_item endpoint). If I set the 'args' value for the delete_item endpoint as an array merge between 'context' and `get_endpoint_args_for_item_schema` it works for me as expected.

```
'args'  => array_merge(
    array(
        'context' => $this->get_context_param(
            array(
                'default' => 'edit',
            )
        ),
    ),
    $this->get_endpoint_args_for_item_schema( WP_REST_Server::DELETABLE )
),
```

renatonascalves:
@sjregan Tks for your feedback.

I updated the code to account for that. A few other changes:

The `get_item`, `update_item`, and `delete_item` requires a user ID now and not a friendship ID as before. This was something discussed in a previous BuddyPress meeting to facilitate the act of fetching, updating and deleting a friendship. Since getting the friendship ID would be harder, but the friend ID or the initiator ID is much more likely to be present when using in an app, for example.

Unit tests updated and improved to account for that. :)

renatonascalves:
@imath I'd appreciate your input here as well. :)

PR #288, opened 2020-03-15
Fixes incorrect error codes returned in get_item requests

sjregan:
The get_item endpoint returns incorrect error codes, this changes the codes to be relevant to the request and to make it clearer why the request failed.

### DIFF
--- a/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
+++ b/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
@@ -209,8 +209,8 @@ class BP_REST_Friends_Endpoint extends WP_REST_Controller {
 		// Check if user is valid.
 		if ( false === $user ) {
 			return new WP_Error(
-				'bp_rest_friends_create_item_failed',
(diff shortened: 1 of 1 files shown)


sjregan:
The problem is that in the get_item endpoint the error is 'bp_rest_friends_create_item_failed'. The endpoint is trying to get the friendship, not create it.

Then if a friendship is not found, the error is bp_rest_invalid_id - 'Invalid friendship ID.' - however a friendship ID wasn't passed in via the query args.

includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
```diff
@@ -224,8 +224,8 @@ public function get_item( $request ) {
 
 		if ( ! $friendship || empty( $friendship->id ) ) {
 			return new WP_Error(
-				'bp_rest_invalid_id',
-				__( 'Invalid friendship ID.', 'buddypress' ),
+				'bp_rest_not_found',
```
renatonascalves:
```suggestion
				'bp_rest_invalid_id',
```

PR #292, opened 2020-03-26
Prepare BuddyPress 6.0.0-beta1 release

imath:
Hi @renatonascalves 

I've reviewed the build process in [#BP7156](https://buddypress.trac.wordpress.org/ticket/7156). The goal of this PR is to make sure everything arrives at the right place into BuddyPress core.

For instance: there's no `bp-signup` directory into BuddyPress. Signups are a Members component feature as you've noticed as you're only loading the controller when the Members component is active. So the REST controller should land into `/bp-members/classes` into BuddyPress. It's the same idea for Group/Member/Blog Avatar or Cover Image.

I know the Group and Member Avatar endpoints landed into `/bp-core/classes` but it was wrong. 

Thanks in advance for reviewing this PR before March 30 so that we can publish BP 6.0.0-beta 1 before the end of march as announced.


imath:
Thanks for your review, let's merge it !

PR #293, opened 2020-03-26
Avoiding regular users from creating friendships to others

### LINKED ISSUE #291: User can create and force friendships for other pairs of users
Currently users can create friendships that do not involve themselves, ie. user 1 can create a friendship between users 2 & 3.

[This pull request](https://github.com/buddypress/BP-REST/pull/290) solves that by forcing the initiator in the friendship to be the one making the API request.

I'm sorry but the pull request does not include updates to the unit tests.

Additionally, the pull request does not address the ability for users to force their friendship upon other users. This is a security hole as plugins and themes sometimes limit which data another user can see unless they are friends.

renatonascalves:
fixes #291

### DIFF
--- a/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
+++ b/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php
@@ -309,8 +309,10 @@ class BP_REST_Friends_Endpoint extends WP_REST_Controller {
 			);
 		}
 
+		$is_moderator = bp_current_user_can( 'bp_moderate' );
+
 		// Only admins can create friendship requests for other people.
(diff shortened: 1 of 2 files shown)


sjregan:
Thanks for working on this.

While this addresses users being able to create friendships that don't involve themselves, it looks to me as though users can still force their friendship upon people through the 'force' parameter, even if they do not have the 'bp_moderate' permission.

renatonascalves:
That's true! Let me check on BuddyPress core if that's an admin only feature. If it is, will update this.

renatonascalves:
@sjregan Code updated to avoid anyone from forcing a friendship request. Instead of returning an error when a "normal" member tries to do that, we are silently creating the friendship request, but not forcing it.

renatonascalves:
@imath Could you take a look when you have a change?

PR #294, opened 2020-03-28
Improve translators comments

imath:
Now we are using WP CLI to generate the BuddyPress pot file, we need to have consistent translators comments for i18n strings. The comment changes are making sure we are using the same in BuddyPress or into the BP REST API considering the same strings. It also fixes WP CLI warnings when releasing BuddyPress.

See [#8260](https://buddypress.trac.wordpress.org/ticket/8260) for more info.


imath:
Thanks for your feedback @renatonascalves Let's merge this :)

PR #296, opened 2020-04-04
Fixing permissions for group mods

renatonascalves:


imath:
FYI, tests are failing because BuddyPress is not using Blogs deprecated hooks. I will open a PR separately to fix failing tests.

PR #298, opened 2020-04-06
Cleaning up

renatonascalves:
**No changes to functionality,** only improving unit tests and removing lots of things that are honestly not needed.


includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
```diff
@@ -514,7 +514,7 @@ public function delete_item_permissions_check( $request ) {
 	}
 
 	/**
-	 * Prepares notification data for return as an object.
+	 * Prepares notification data tp return as an object.
```
imath, line 517: What is "tp" ? It's probably a typo, you should have meant WP or BP :)

renatonascalves:
@imath Added your code review suggestions. :)

PR #299, opened 2020-04-17
Stop forcing the context parameter of the request

imath:
Since [WP47559](https://core.trac.wordpress.org/changeset/47559/) forcing this parameter generates failing tests.

I do not think we need to do this.

Could you have a look at it @renatonascalves ?


imath:
Thanks for your feedback 👍

PR #300, opened 2020-04-17
Make sure only the Group Admin can delete others membership

imath:
#299 is fixing the failing tests here.

I've checked, `groups_remove_member()` returns false if `!  bp_is_item_admin()`. `bp_is_item_admin()` is not containing mods but only admins, see [src/bp-groups/classes/class-bp-groups-component.php#L311](https://github.com/buddypress/BuddyPress/blob/master/src/bp-groups/classes/class-bp-groups-component.php#L311)


imath:
NB: this PR will need to be backport on 0.1 branch.

imath:
Thanks for your feedback 👍

PR #317, opened 2020-05-16
Group Membership Endpoint: search_terms should default to false

imath:
See https://buddypress.trac.wordpress.org/ticket/8299

In short using `NULL` as the default value can generate SQL overheat unnecessarily.

BTW we should probably create a 0.2 branch in case some new features need to be added to the plugin.

This way this 0.2 branch would be used by the BuddyPress 6.0 branch.


imath:
Thanks for your review @renatonascalves 👍. Let's have this in. I'll take care of creating the 0.2 branch just after.

PR #325, opened 2020-06-14
Add a groups/me route to get current users’s group

imath:
In order to easily fetch current user’s groups in JavaScript, it would be very convenient to add this route. I believe, it might also be interesting for some other components (friends, blogs..).

This groups route is really important when checking if a user can post activities to groups for instance. (currently working on using the BP REST API to post activities 😉 )

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -87,6 +87,28 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 				'schema' => array( $this, 'get_item_schema' ),
 			)
 		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/me',
+			array(
+				'args'   => array(
+					'max' => array(
+						'description' => __( 'The maximum amount of groups the user is member of to return. Defaults to all groups.', 'buddypress' ),
+						'type'        => 'integer',
+						'default'     => 0,
+					),
+				),
+				array(
+					'methods'  => WP_REST_Server::READABLE,
+					'callback' => array( $this, 'get_current_user_groups' ),
+					'args'     => array(
+						'context' => $this->get_context_param( array( 'default' => 'view' ) ),
+					),
+				),
+				'schema' => array( $this, 'get_item_schema' ),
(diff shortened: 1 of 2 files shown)


includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -87,6 +87,28 @@ public function register_routes() {
...
+				'args'   => array(
+					'max' => array(
+						'description' => __( 'The maximum amount of groups the user is member of to return. Defaults to all groups.', 'buddypress' ),
+						'type'        => 'integer',
+						'default'     => 0,
+					),
+				),
+				array(
+					'methods'  => WP_REST_Server::READABLE,
+					'callback' => array( $this, 'get_current_user_groups' ),
```
renatonascalves, line 104: I think it is missing the `permission_callback`. This would be required since there is a ticket somewhere where the REST API team is considering making it a required.

So I'd recommend adding here.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -583,6 +605,80 @@ public function delete_item_permissions_check( $request ) {
...
+
+		if ( empty( $current_user_id ) ) {
+			return new WP_Error(
+				'bp_rest_authorization_required',
+				__( 'Sorry, you need to be logged in to view your groups.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		}
```
renatonascalves, line 628: This would be added inside the `permission_callback`...

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
imath: Hi @renatonascalves 

Fine with me. Thanks for your review. I've just added the permission checks function in this commit https://github.com/buddypress/BP-REST/pull/325/commits/9c3ef5dafb91a035a697a46457750d9ad44389d1

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
imath: Got it 👍 Thanks.

PR #327, opened 2020-06-25
Allow Group details to be available in its single endpoint

### LINKED ISSUE #314: Permission checks when getting private group details
This isn't so much an issue with the code but a question/discussion around the handling of permissions when getting group details via the get_items endpoint vs the get_item endpoint.

If I query the get_items endpoint (`/wp-json/buddypress/v1/groups`) while unauthenticated the response is:

```
{
  "id": 88,
  "creator_id": 922,
  "parent_id": 0,
  "date_created": "2020-05-12T05:04:41",
  "description": {
    "raw": "Private group desc.",
    "rendered": "<p>Private group desc.</p>\n"
  },
  "enable_forum": false,
  "link": "http://domain.test/groups/private-group/",
  "name": "Private group",
  "slug": "private-group",
  "status": "private",
  "types": [],
  "avatar_urls": {
    "full": "http://domain.test/wp-content/plugins/buddypress/bp-core/images/mystery-group.png",
    "thumb": "http://domain.test/wp-content/plugins/buddypress/bp-core/images/mystery-group-50.png"
  },
  "_links": {
    "self": [
      {
        "href": "http://domain.test/wp-json/buddypress/v1/groups/88"
      }
    ],
    "collection": [
      {
        "href": "http://domain.test/wp-json/buddypress/v1/groups/"
      }
    ],
    "user": [
      {
        "embeddable": true,
        "href": "http://domain.test/wp-json/buddypress/v1/members/922"
      }
    ]
  }
}
```

If I query the get_item endpoint (`/wp-json/buddypress/v1/groups/88`) while unauthenticated the response is:

```
{
  "code": "bp_rest_authorization_required",
  "message": "Sorry, you cannot view the group.",
  "data": {
    "status": 401
  }
}
```

This means through the list endpoint I can get private group details such as the creator, parent, date created and most importantly name, slug, description and avatar URLs whereas through the single group endpoint I cannot get any information regarding the group at all.

In my use case it would be really useful to still be able to get the same information in the single endpoint that you can in the list endpoint - the information is already exposed to an unauthenticated user.

Should the single endpoint be exposing the data and this is an issue?

renatonascalves:
The information from the `buddypress/v1/groups/group_id` endpoint is already available in the `get_items` endpoint from this component. So there is no need to hide it from public view.

Also, with this change, a few unit tests became redudant, so they were removed.

closes #314

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -272,16 +272,6 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 			);
 		}
 
-		if ( true === $retval && ! $this->can_see( $group ) ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you cannot view the group.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		}
-
 		/**
 		 * Filter the groups `get_item` permissions check.
 		 *
@@ -966,34 +956,6 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 	}
 
 	/**
-	 * Can a user see a group?
-	 *
-	 * @since 0.1.0
-	 *
-	 * @param  BP_Groups_Group $group Group object.
-	 * @return bool
-	 */
-	protected function can_see( $group ) {
-
-		// If it is not a hidden/private group, user can see it.
-		if ( 'public' === $group->status ) {
-			return true;
-		}
-
-		// Moderators.
-		if ( bp_current_user_can( 'bp_moderate' ) ) {
-			return true;
-		}
-
-		// User is a member of the group.
-		if ( groups_is_user_member( bp_loggedin_user_id(), $group->id ) ) {
-			return true;
-		}
-
-		return false;
-	}
-
-	/**
 	 * See if user can delete or update a group.
 	 *
 	 * @since 0.1.0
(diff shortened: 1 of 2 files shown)


includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -1020,11 +982,11 @@ protected function can_see_hidden_groups( $request ) {
...
 
-			if ( is_user_logged_in() && isset( $request['user_id'] ) && absint( $request['user_id'] ) === bp_loggedin_user_id() ) {
-				return true;
-			}
-
-			return false;
+			return (
+				is_user_logged_in()
+				&& isset( $request['user_id'] )
+				&& absint( $request['user_id'] ) === bp_loggedin_user_id()
```
imath, line 988: Unlike for private groups, hidden groups are only listed when you are a member of it or when you have the `bp_moderate' capability. So I believe you need an extra `groups_is_user_member()` check here.

tests/groups/test-controller.php
```diff
@@ -243,66 +197,13 @@ public function test_get_private_group() {
 	/**
 	 * @group get_item
 	 */
-	public function test_get_hidden_group_without_being_from_group() {
```
imath, line 246: This should still being tested.

tests/groups/test-controller.php
```diff
@@ -179,55 +179,9 @@ public function test_get_item() {
...
-		$request  = new WP_REST_Request( 'GET', sprintf( $this->endpoint_url . '/%d', REST_TESTS_IMPOSSIBLY_HIGH_NUMBER ) );
-		$response = $this->server->dispatch( $request );
-
-		$this->assertErrorResponse( 'bp_rest_group_invalid_id', $response, 404 );
-	}
-
-	/**
-	 * @group get_item
-	 */
-	public function test_get_hidden_group() {
```
imath, line 194: This should still be tested.

renatonascalves:
@imath I've addressed your suggestions! Could you take another look?

PR #331, opened 2020-07-07
List hidden groups if user has permissions, otherwise return public and private groups

renatonascalves:
I'm suggesting a change in how we check for the permissions when trying to list groups. Right now, if a person tries to list hidden groups but have no permission to do so, we return an error object. That seems too much.

I'm suggesting we don't show an error object but instead return the public and private groups. Users with permissions to see the groups, will continue to do so if they choose it, of course.

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -144,7 +144,7 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 		}
 
 		// See if the user can see hidden groups.
-		if ( isset( $request['show_hidden'] ) && true === (bool) $request['show_hidden']  && ! $this->can_see_hidden_groups( $request ) ) {
+		if ( isset( $request['show_hidden'] ) && true === (bool) $request['show_hidden'] && ! $this->can_see_hidden_groups( $request ) ) {
 			$args['show_hidden'] = false;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -143,6 +143,10 @@ public function get_items( $request ) {
 			$args['parent_id'] = null;
 		}
 
+		if ( isset( $request['show_hidden'] ) && ! $this->can_see_hidden_groups( $request ) ) {
```
imath, line 146: I'd suggest to check `$request['show_hidden']` is true. Something like that:

```php
if ( isset( $request['show_hidden'] ) && true === (bool) $request['show_hidden']  && ! $this->can_see_hidden_groups( $request ) ) {
```

renatonascalves: Yeah, that makes total sense. Will update. :)

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -143,6 +143,11 @@ public function get_items( $request ) {
 			$args['parent_id'] = null;
 		}
 
+		// See if the user can see hidden groups.
+		if ( isset( $request['show_hidden'] ) && true === (bool) $request['show_hidden']  && ! $this->can_see_hidden_groups( $request ) ) {
```
renatonascalves, line 147:
```suggestion
		if ( isset( $request['show_hidden'] ) && true === (bool) $request['show_hidden'] && ! $this->can_see_hidden_groups( $request ) ) {
```

PR #335, opened 2020-07-09
Make sure the response of activity posted in a group is the activity

imath:
When posting an activity into a private or hidden group, the response does not include the posted activity because in this case `$request['hidden']` is not set.

Using `true` for the `show_hidden` argument used into `bp_activity_get` seems fine as anyway the activity has been created by the user, I don't see any reason preventing him to be able to read it 😃 

I've added some tests for the `create_item` method and one for the `update_item` method to make sure this issue wasn't also appearing when updating an activity of a non public group.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -421,7 +421,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			array(
 				'in'               => $activity_id,
 				'display_comments' => 'stream',
-				'show_hidden'      => $request['hidden'],
+				'show_hidden'      => true,
 			)
 		);
 
--- a/tests/activity/test-controller.php
+++ b/tests/activity/test-controller.php
@@ -302,7 +302,7 @@ class BP_Test_REST_Activity_Endpoint extends WP_Test_REST_Controller_Testcase {
 		$this->assertEquals( 200, $response->get_status() );
 
 		$a_ids = wp_list_pluck( $response->get_data(), 'id' );
-		
+
 		$this->assertContains( $a1, $a_ids );
 		$this->assertContains( $a2, $a_ids );
 	}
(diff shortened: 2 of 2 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -421,7 +421,7 @@ public function create_item( $request ) {
 			array(
 				'in'               => $activity_id,
 				'display_comments' => 'stream',
-				'show_hidden'      => $request['hidden'],
+				'show_hidden'      => true,
```
renatonascalves, line 424: I'd suggest another change in this file. 

* Removing the `hidden` parameter from the `POST/PUT` schema at `get_endpoint_args_for_item_schema`

tests/activity/test-controller.php
```diff
@@ -559,6 +559,56 @@ public function test_create_item_in_a_group() {
 		$this->check_create_activity_response( $response );
 	}
 
+	/**
+	 * @group create_item
```
renatonascalves, line 563: There is a unit test `test_create_blog_post_item` that's using this parameter:

```
$params = $this->set_activity_data( array(
	'component'         => buddypress()->blogs->id,
	'primary_item_id'   => get_current_blog_id(),
	'secondary_item_id' => $p,
	'type'              => 'new_blog_post',
	'hidden'            => true,
) );
```

Could you remove that as well?

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
imath: Well, I believe we should probably left it as you can always create a regular hidden activity or update a public activity to be hidden.

tests/activity/test-controller.php
imath: Well, considering my first comment, I believe it's a way to test it's possible to create an activity as hidden 😉

imath:
@renatonascalves what's your opinion about the comments I made about the changes you requested?

renatonascalves:
@imath Disregard my comments. I mixed things up. This is good to go!

imath:
Thanks a lot for your feedback @renatonascalves I'm merging the PR 😉

PR #336, opened 2020-07-10
Update wrong HTTP status codes

adiloztaser:
Hey team :hand:,

We should return `4xx` error code instead of `500` for bad requests like `empty_content`, `invalid_id` etc.  A `500` error code is for internal server errors.

I updated some of them and I'll keep update but I'm open to help and any suggestions.


adiloztaser:
Thanks for your comments. I agree both of you, we need to inform developers about the changes and development post is a good way to do it.

For now I am converting the PR to draft. I'll keep to improvements for other endpoints.

adiloztaser:
I've updated all wrong response statuses which I found and I believe it's ready to review.

PR #337, opened 2020-07-11
Single member item permission checks

### LINKED ISSUE #326: Permission checks when getting single member item
This is a continuation of issue #314 - although rather than relating to the groups endpoint it relates to the members endpoint.

An unauthenticated user, using the `get_items` endpoint can retrieve member details such as:

- name
- user_login
- link
- member type
- xprofile data
- mention_name
- avatar URLs
- slug

However if an unauthenticated user uses the `/members/<id>` endpoint a `bp_rest_authorization_required` error is returned.

This creates odd situations such as an unauthenticated user being able to see the activity of a member, but not the member themselves.

renatonascalves:
fixes #326

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -136,20 +136,9 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 	 */
 	public function get_item_permissions_check( $request ) {
 		$retval = true;
+		$user   = bp_rest_get_user( $request['id'] );
 
-		if ( ! is_user_logged_in() ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you are not allowed to view members.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		}
-
-		$user = bp_rest_get_user( $request['id'] );
-
-		if ( true === $retval && ! $user instanceof WP_User ) {
+		if ( ! $user instanceof WP_User ) {
 			$retval = new WP_Error(
 				'bp_rest_member_invalid_id',
 				__( 'Invalid member ID.', 'buddypress' ),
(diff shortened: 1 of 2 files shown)


includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -391,7 +380,7 @@ public function user_data( $user, $context = 'view' ) {
 			$data['friendship_status']      = ( 'is_friend' === $friendship_status );
 		}
 
-		if ( 'edit' === $context ) {
+		if ( 'edit' === $context && current_user_can( 'list_users' ) ) {
```
renatonascalves, line 383: This was added so that whoever is extending or using this method, they don't share sensitive information to the public.

Right now, we are using this method in the Group Membership endpoint.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -136,20 +136,9 @@ public function get_items_permissions_check( $request ) {
...
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		}
-
-		$user = bp_rest_get_user( $request['id'] );
-
-		if ( true === $retval && ! $user instanceof WP_User ) {
+		if ( ! $user instanceof WP_User ) {
```
imath, line 141: I wonder if a not logged in user is not a WP_User. I believe it's probably the case: i would probably use `current_user_can( 'exist' )` instead.

renatonascalves:
@imath Could you elaborate a bit more? I think I'm not getting your point. :/

The goal is allow anyone to see the user information via the single member endpoint. You don't need to be logged in to see it.

The user ID provided is checked to make sure we are returning valid user information. Otherwise, no need to go any further.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Oops, my bad, sorry I thought it was an access check. @renatonascalves Ignore my comment 😅

PR #346, opened 2020-07-26
Get activities using more than one type

### LINKED ISSUE #345: List activities from multiple types
I am trying to get activities from multiple types by adding two activity types on the `type` parameter

```
type: 'activity_update,new_blog_post',
```

I am using the below code:

```
bp.apiRequest( {
  path: 'buddypress/v1/activity',
  type: 'GET',
  data: {
    context: 'view',
    type: 'activity_update,new_blog_post',
  }
} ).done( function( data ) {
  return data;
} ).fail( function( error ) {
  return error;
} );
```

The response to that above call is:

```
{
    "code": "rest_invalid_param",
    "message": "Invalid parameter(s): type",
    "data": {
        "status": 400,
        "params": {
            "type": "type is not one of new_member, new_avatar, updated_profile, activity_update, activity_comment, friendship_accepted, friendship_created, created_group, joined_group, group_details_updated, new_blog_post, new_blog_comment."
        }
    }
}
```

Is this something that is not supported, or am I doing something wrong? I couldn't find any reference on the documentation.

Thank you in advance.

renatonascalves:
fixes #345

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1546,9 +1546,12 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 
 		$params['type'] = array(
 			'description'       => __( 'Limit result set to items with a specific activity type.', 'buddypress' ),
-			'type'              => 'string',
-			'enum'              => array_keys( bp_activity_get_types() ),
-			'sanitize_callback' => 'sanitize_key',
+			'type'              => 'array',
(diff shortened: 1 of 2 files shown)


renatonascalves:
@imath do you have any objection of me adding this in the `functions.php` file here rather than in BuddyPress core?

imath:
@renatonascalves well, I can add it myself if you want. I think it's best to directly do it in BP, because we won't have to think about not forgetting to do it when we'll prepare the 7.0.0 release.
I'll ping you from this PR once it's committed in BP 😉

PR #348, opened 2020-07-26
Attachments fix for those on Windows/WAMP installs

renatonascalves:
https://buddypress.trac.wordpress.org/ticket/8336


renatonascalves:
Yeah, tks for confirming that as well.

PR #355, opened 2020-09-27
Include BuddyPress specific extra data to members

imath:
I think we should Introduce a new collection parameter for the members::get_items() endpoint `populate_extras` which if set to true would include into the endpoint's response:

- the last time the each user were active,
- the latest activity update of each user,
- the number of friends each user has.

1. This would help me to use these specific pieces of data to build the BP Member**s** block I’m working on.
2. I also believe the `members::get_item()` endpoint should include these pieces of data so that we could update the item schema to include these.


imath:
FYI, about point 2. I've created this [ticket ](https://buddypress.trac.wordpress.org/ticket/8371)in BP Trac to discuss about it with the team.

PR #357, opened 2020-10-06
Use a `populate_extras` query arg to get group/s number of members & last activity

imath:
Working on the BP Groups block, I needed to get some extra information about the group. As the number of members and the last activity are displayed on the front end publicly I believe we can safely fetch these pieces of information even if not into the edit context.


renatonascalves (approved):
All good!

PR #358, opened 2020-11-01
Take in account BP 7.0.0 changes about default blavatar

imath:
If the `no_user_grav` param is set to true, then the blavatar fetched will be the default one.
Add unit tests to check the 3 available scenario when fetching the blog avatar:
- site icon is used
- no user gravatar is requested
- use the admin user avatar

See: https://buddypress.trac.wordpress.org/changeset/12772


renatonascalves (approved):
Looks ok!

imath:
Thanks for your feedback @renatonascalves 😍

PR #362, opened 2020-11-07
Improve Members endpoints schemas

imath:
These are small fixes to make sure schemas are consistent. The Members Endpoint documentation has been updated accordingly, see https://developer.buddypress.org/bp-rest-api/reference/members/

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -799,6 +799,7 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 					'arg_options' => array(
 						'sanitize_callback' => 'sanitize_text_field',
 					),
+					'readonly'    => true,
 				),
 				'link'               => array(
 					'description' => __( 'Profile URL of the member.', 'buddypress' ),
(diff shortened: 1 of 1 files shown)


includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -905,7 +907,7 @@ public function get_item_schema() {
 						'raw'      => array(
 							'description' => __( 'Content for the activity, as it exists in the database.', 'buddypress' ),
 							'type'        => 'string',
-							'context'     => array( 'edit' ),
+							'context'     => array( 'view', 'edit' ),
```
renatonascalves, line 910: I thought `last_update` was not publicly viewable.

renatonascalves (approved):

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Hi @renatonascalves 
Thanks for your comment & PR approval, I've done some searches. I confirm the `latest_update` is publicly viewable. It's saved as a user meta when using `bp_activity_post_update()` which is only used from the user's or directory activity post form on front end, which means it cannot be hidden. It's displayed into the Member's header to every site visitor as long as the activity component is active.

PR #372, opened 2020-12-11
0.4.0 version bumps

imath:
From now on we need to use `@since 8.0.0` for new introduced functions/classes as this will be included into BuddyPress 8.0.0. FYI, the 0.3 branch has been created for the BuddyPress 7.0 branch


renatonascalves (approved):
All good!

imath:
Thanks for your review 👌

PR #380, opened 2021-03-05
Members: Preserve order of XProfile groups/fields

### LINKED ISSUE #375: xprofile_data function doesn't preserve the order of fields/groups when converted to json
When fetching the member endpoint the profile data is not In the same order as the admin UI because the keys are all integers and JSON orders integer keys sequentially. xprofile_data function is using the field id as the key but it should be the field_order / group_order as the key to keep the data in order. 

```	
    protected function xprofile_data( $user_id ) {
		$data = array();

		// Get XProfile groups, only if the component is active.
		if ( bp_is_active( 'xprofile' ) ) {
			$fields_endpoint = new BP_REST_XProfile_Fields_Endpoint();

			$groups = bp_xprofile_get_groups(
				array(
					'user_id'          => $user_id,
					'fetch_fields'     => true,
					'fetch_field_data' => true,
				)
			);

			foreach ( $groups as $group ) {
				$data['groups'][ $group->id ] = array(
					'id'   => $group->id,
					'name' => $group->name,
				);

				foreach ( $group->fields as $item ) {
					$data['groups'][ $item->group_id ]['fields'][ $item->field_order ] = array(
						'id'    => $item->id,
						'name'  => $item->name,
						'value' => array(
							'raw'          => $item->data->value,
							'unserialized' => $fields_endpoint->get_profile_field_unserialized_value( $item->data->value ),
							'rendered'     => $fields_endpoint->get_profile_field_rendered_value( $item->data->value, $item ),
						),
					);
				}
			}
		} else {
			$data = array( __( 'No extended profile data available as the component is inactive', 'buddypress' ) );
		}

		return $data;
	}
```

renatonascalves:
fixes #375

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /node_modules
 /npm-debug.log
 /vendor/
+.vscode
 
 # Operating system specific files
 .DS_Store
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -636,14 +636,16 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 				)
 			);
 
-			foreach ( $groups as $group ) {
-				$data['groups'][ $group->id ] = array(
+			foreach ( $groups as $group_order => $group ) {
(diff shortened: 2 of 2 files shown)


imath (approved):
Sure!

PR #386, opened 2021-03-24
Caching schema in REST controllers

### LINKED ISSUE #225: Cache item schema in REST controllers
Interesting ticket discussing performance improvements: https://core.trac.wordpress.org/ticket/47871

We will need to do the same or something similar to improve the performance on our endpoints/controllers.

renatonascalves:
fixes #225 

- Cache schema from REST controllers
- Update schema cache hook param
-  Unit tests are running on all endpoints on multisite
- Skipped some tests on multisite (they require more investigation)
- Removing the `embed` context for consistency
- Added unit test for a member without an avatar
- updated `@wordpress/env` from its temporary url to the latest version: 4.0

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1415,12 +1415,17 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			);
 		}
 
+		// Cache current schema here.
+		if ( is_null( $this->schema ) ) {
+			$this->schema = $schema;
+		}
+
 		/**
 		 * Filters the activity schema.
 		 *
-		 * @param string $schema The endpoint schema.
+		 * @param array $schema The endpoint schema.
 		 */
-		return apply_filters( 'bp_rest_activity_schema', $this->add_additional_fields_schema( $schema ) );
+		return apply_filters( 'bp_rest_activity_schema', $this->add_additional_fields_schema( $this->schema ) );
 	}
 
 	/**
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -929,7 +929,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			}
 		}
 
-		if ( ! empty( $schema['properties']['user_avatar'] ) ) {
+		if ( true === buddypress()->avatar->show_avatars ) {
 			$data['user_avatar'] = array(
 				'full'  => bp_core_fetch_avatar(
 					array(
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1262,161 +1262,160 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 * @return array
 	 */
 	public function get_item_schema() {
-		$schema = array(
-			'$schema'    => 'http://json-schema.org/draft-04/schema#',
-			'title'      => 'bp_activity',
-			'type'       => 'object',
-			'properties' => array(
-				'id'                => array(
-					'context'     => array( 'view', 'edit' ),
-					'description' => __( 'A unique numeric ID for the activity.', 'buddypress' ),
-					'readonly'    => true,
-					'type'        => 'integer',
-				),
-				'primary_item_id'   => array(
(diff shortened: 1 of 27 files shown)


tests/multisite.xml
```diff
@@ -11,8 +11,7 @@
 	</php>
 	<testsuites>
 		<testsuite name="unit">
-			<directory prefix="test-" suffix=".php">./blogs/</directory>
-			<directory prefix="test-" suffix=".php">./attachments/</directory>
+			<directory prefix="test-" suffix=".php">./</directory>
```
renatonascalves, line 14: We should be testing everything on multisite.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -1291,12 +1291,17 @@ public function get_item_schema() {
 			);
 		}
 
+		// (Re)cache current schema here.
```
imath, line 1294: Looking at this https://core.trac.wordpress.org/changeset/45811 I don't see any cache related functions so I assume the $this->schema is cleared each time a request is made no ? I believe the object cache is made to avoid setting the schema more than once per request, so I doubt the avatar upload setting would change during a request.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -962,26 +963,34 @@ public function get_item_schema() {
...
 				'type'        => 'object',
-				'context'     => array( 'embed', 'view', 'edit' ),
+				'context'     => array( 'view', 'edit' ),
 				'readonly'    => true,
 				'properties'  => $avatar_properties,
 			);
 		}
 
+		// Update schema if member types don't match.
+		$update_schema = ( $this->schema && $schema['properties']['member_types']['enum'] !== $this->schema['properties']['member_types']['enum'] );
```
imath, line 987: I'm unsure this is needed, for the same reason than my previous comment :)

imath (commented):
Hi @renatonascalves 

Thanks a lot for your work on this. I believe to really benefit from this, the schema should only be set if not already the case. I believe the `$schema` is always set and the `$this->schema` is only set to `$schema` if not already the case.

+ I'm unsure some checks are needed as I don't believe the cache persist once the request is finished.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
renatonascalves: Really odd! I was having issues before with this schema but after trying out again, and removing this line, it worked correctly. Dunno what happened but this is not needed.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
renatonascalves: The cached schema was coming from the `prepare_item_for_response` method. This fix is not longer needed. :)

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -1415,12 +1415,17 @@ public function get_item_schema() {
 			);
```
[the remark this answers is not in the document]
renatonascalves: The end result is the same using the current way or using this way. Both will cache the schema only if it is not already.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -1415,12 +1415,17 @@ public function get_item_schema() {
...
 		}
 
+		// Cache current schema here.
+		if ( is_null( $this->schema ) ) {
+			$this->schema = $schema;
+		}
+
```
imath, line 1422: So my point was doing it this way the variable `$schema` is always set even if `$this->schema` is set. So all the above code is always run. We could avoid this getting rid of `$schema` and setting the schema directly in `$this->schema`.

imath (commented):
If you look at how WordPress does you'll see the code to set the schema is only run if not cached. Doing it the way you chose, you always run the code to set the schema no matter it was cached or not. That seems a bit weird to me, have you tried to check the way I suggested in my first comment to see if it was improving performance more ?

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
renatonascalves: That's a very good point. I'm wondering now how I'm getting good performances though.

renatonascalves:
@imath I updated to reflect your suggestion.

imath (approved):
Thanks a lot for your work 💪. Let's have this in 👌

PR #388, opened 2021-04-01
Decrease number of requests to get the schema

### LINKED ISSUE #387: Decrease number of requests to get the schema
Fun fact: in a request to http://site.com/wp-json/buddypress/v1/activity?per_page=100 the `get_item_schema` method is called 505 times.

The use of `$schema = $this->get_item_schema();` inside `prepare_item_for_response` methods is the culprit.

renatonascalves:
fixes #387

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -916,15 +916,12 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			'favorited'         => in_array( $activity->id, $this->get_user_favorites(), true ),
 		);
 
-		// Get item schema.
-		$schema = $this->get_item_schema();
-
 		// Get comments (count).
 		if ( ! empty( $activity->children ) ) {
 			$comment_count         = wp_filter_object_list( $activity->children, array( 'type' => 'activity_comment' ), 'AND', 'id' );
 			$data['comment_count'] = ! empty( $comment_count ) ? count( $comment_count ) : 0;
 
-			if ( ! empty( $schema['properties']['comments'] ) && 'threaded' === $request['display_comments'] ) {
+			if ( ! empty( $this->schema['properties']['comments'] ) && 'threaded' === $request->get_param( 'display_comments' ) ) {
 				$data['comments'] = $this->prepare_activity_comments( $activity->children, $request );
 			}
 		}
@@ -1011,7 +1008,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 	 */
 	protected function prepare_item_for_database( $request ) {
 		$prepared_activity = new stdClass();
-		$schema            = $this->get_item_schema();
+		$schema            = $this->schema;
 		$activity          = $this->get_activity_object( $request );
 
 		if ( ! empty( $schema['properties']['id'] ) && ! empty( $activity->id ) ) {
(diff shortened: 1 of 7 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -916,15 +916,12 @@ public function prepare_item_for_response( $activity, $request ) {
 			'favorited'         => in_array( $activity->id, $this->get_user_favorites(), true ),
 		);
 
-		// Get item schema.
-		$schema = $this->get_item_schema();
```
imath, line 920: Now that you have committed the improvement about caching the schema, are you sure we need to remove this and use $this->schema instead.  $this->schema should be the same as $this->get_item_schema() and it shouldn't run extra code unless $this->schema is null see https://github.com/buddypress/BP-REST/blob/9fe7c369f2dfe713a6e8974cdab4e6603ffee6b0/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php#L1262 Or am I missing something?

The risk to not use $this->get_item_schema() is to miss additional fields as there are not into $this->schema but into $this->get_item_schema()

imath (commented):
I must say I'm a bit dubitative 🤔  I believe it's now fine to use `$this->get_item_schema()` now you committed the cache improvements and I thought these improvements were made to be able to do so.

includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
renatonascalves: There is an increase in performance in doing that. Mainly because of the number of times that method is run each time when generating a response.

> The risk to not use $this->get_item_schema() is to miss additional fields as there are not into $this->schema but into $this->get_item_schema()

Since the schema is run before the `prepare_item_for_response`  method, I don't know how that would be possible. The schema cache doesn't change when looping an array of items to return.

imath: Could you check again? I believe the way you did it `$this->schema` never includes the additional fields because I believe the filter is only run at the bottom of the `get_item_schema()` method on the cached schema. Meaning If you use the cached schema elsewhere you never get potential additional fields.

You could alternatively cache the additional fields into the schema moving the filter to the place you set `$this->schema = $schema`.

My fear is some plugins might lack their needed additional fields where you directly use `$this->schema` according to the way the PR is currently built.

renatonascalves:
@imath 

Since the `get_item_schema` is being used to confirm if some endpoint schema fields are present, not additional fields added by plugins, I think this is not an issue here.

In the cases where `get_item_schema`  is being used inside `prepare_item_for_database`, for example, we are not passing that schema to the hook. And additional fields are not being verified inside that method as well. Again, not an issue here in fetching from cache.

> You could alternatively cache the additional fields into the schema moving the filter to the place you set $this-> schema = $schema.

Here is a good and valid reason of why not to do this: https://core.trac.wordpress.org/ticket/47871#comment:5 This is why I opted not to cache that and followed WordPress core.

> My fear is some plugins might lack their needed additional fields where you directly use $this->schema according to the way the PR is currently built.

I'm not sure how that would happen. Again: we are not passing the schema to any hook when using them inside `prepare_item_for_database` or `prepare_item_for_response`. Schema hooks like `bp_rest_activity_schema` are being sent with the additional fields, so plugins will continue to access them without this change affecting them.

:)

imath (approved):
Ok, these seem to be good points. I must say, to be safer, I'd rather keep on using the `get_item_schema()` now it caches the schema just like [WordPress does](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1087) 🤗

renatonascalves:
Sounds good!

PR #391, opened 2021-05-11
Include Group id in the group Members response

sbrajesh:
Group id should be included in the response for the group membership endpoint. This is missing in the current version. This patch adds it.

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -707,6 +707,7 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 		$data = array_merge(
 			$member_data,
 			array(
+				'group_id'      => (int) $group_member->group_id,
 				'is_mod'        => (bool) $group_member->is_mod,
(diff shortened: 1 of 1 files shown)


imath (commented):
Thanks for your PR @sbrajesh I agree it can be useful 👌. @renatonascalves can you look at it ? There might be some edits to do in unit tests to take the in account the new schema prop.

sbrajesh:
Thank you @imath  I am looking forward to @renatonascalves review.

renatonascalves:
I think I'm going to close this issue and open a new one with a more broader patch.

The unit tests are failing because in some instances, that I need to investigate, that `prepare_item_for_response` method passes a member object with no groupd ID in it. However it should have, so I'm not sure what's happening. 

I need to investigate it further.

--

Also, I don't think this should be an ID in the response. This should be a `embeddable` group in the `prepare_links`. So that more information about the group can be grathered if necessary.

More info here: https://developer.wordpress.org/rest-api/using-the-rest-api/linking-and-embedding/#embedding

But thank you for the patch @sbrajesh I'm gonna try to put something together in the next couple of days. Pull requests are welcomed as well. :)

renatonascalves:
Closing this one in favor of #394

PR #392, opened 2021-05-12
Avoid SQL overheat into the Members get_items endpoint

imath:
See https://buddypress.trac.wordpress.org/ticket/8472

If no search terms were requested, we need to set the argument to false.


renatonascalves (approved):
🗡️

PR #394, opened 2021-05-22
Reviewing Components Prepare Links

### LINKED ISSUE #393: Review Component Prepare Links
Review if components with associated objects can have their information added as links using the `prepare_links`.

Also, review the links themselves. This, https://github.com/buddypress/BP-REST/blob/4856cc74be2f32612e852f832d6b69997ce1b6a7/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php#L1130, for example is wrong. My mistake.

Note: https://developer.wordpress.org/rest-api/using-the-rest-api/linking-and-embedding/#embedding

Related to #391

renatonascalves:
fixes #393 

- Components were reviewed so that their `prepare_links` were correct
- Missing objects were added
- Some had their keys changed to better reflect what they are
- Embedable componets had their schema context updated (@imath it's important to review which fields **should not** be available when embedding an object)

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1123,11 +1123,32 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 			);
 		}
 
+		// Embed Group.
 		if ( bp_is_active( 'groups' ) && 'groups' === $activity->component && ! empty( $activity->item_id ) ) {
-			$group = groups_get_group( $activity->item_id );
-
 			$links['group'] = array(
-				'href'       => bp_get_group_permalink( $group ),
+				'href'       => rest_url(
+					sprintf(
+						'/%s/%s/%d',
+						$this->namespace,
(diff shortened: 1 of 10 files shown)


imath (approved):
Hi @renatonascalves 

It looks good to me. Anyway, about activity and groups: if items are hidden, and the current user has not access to them, as these items are not fetched the corresponding embed links are not either, if this is correct, then it's fine, otherwise, then hidden activity/groups shouldn't be embeddables.

renatonascalves:
@imath That's correct! I tested that locally. But I'll put some unit tests together, something I should have done in the first place. :)

About the fields, do you have any fields you think should be?

imath:
imho Fields are ok to be embed. The most sensitive part is the user's email, and it's not part of them 👌

renatonascalves:
I'll be merging that. I added two changes:

- Added some simple unit tests just to confirm embeddable objects are not available for those without permission.
- I reverted the change in the embeddable keys. This was a breaking change for apps already using those keys, I'm glad I noticed before merging it.

:)

PR #395, opened 2021-05-29
Blogs: Embed the latest blog post

renatonascalves:
Related to #393 

- the latest blog id was added to the response
- one can embed the latest post from another site/blog in the network in the Blog response 🎈 
- Blog description was updated to return the `raw` and `rendered` values (**this is a minor but a breaking change**)


bp-rest.php
```diff
@@ -178,3 +178,106 @@ function() {
...
+	// Request route and requested URL path should match.
+	if ( $request->get_route() !== $bits['path'] ) {
+		return $response;
+	}
+
+	// Get blog id.
+	// @todo update with a cached version.
+	// @todo send a patch to BuddyPress core with a way to fetch a single blog object directly,
+	// with a blog_id or domain.
+	$blog_id = $wpdb->get_var( // phpcs:ignore
```
imath, line 254: What about using [get_site_by_path()](https://developer.wordpress.org/reference/functions/get_site_by_path/)?

imath (approved):
Interesting 👌 Thanks for working on it. I'd advise to add tests when subdomains are used instead of subdirectories 😅

bp-rest.php
renatonascalves: That's interesting! Didn't think of that. Thanks for bring that to my attention. :)

renatonascalves:
@imath Could you code review again?

imath (approved):
Looks good! Thanks for updating the PR 👍

PR #399, opened 2021-06-15
Notifications: notification does not return an invalid object

### LINKED ISSUE #397: Notification delete bug
The documentation says to send DELETE request to: `buddypress/v1/notifications/<id>`

But you actually have to pass an id param as well `buddypress/v1/notifications/<id>?id=<id>`

If you just send `buddypress/v1/notifications/<id>` it will return true like it deleted the notification but it doesn't really delete as the id param is missing.

code needs a better check if param is missing and send a did not delete response.

https://developer.buddypress.org/bp-rest-api/reference/screen-notifications/#definition-5

renatonascalves:
see #397


imath (approved):
Looks good 👍 Thanks for improving this part 👌

PR #402, opened 2021-06-16
Activity: updating controller to fetch params properly

renatonascalves:
as titled.


imath (approved):
Thanks for improving this part, looks good!

PR #403, opened 2021-06-16
Remove the legacy functions.php file with REST API helpers

renatonascalves:
The functions present in that file are loaded from BuddyPress core and the plugin doesn't work by itself, so they are unnecessary, hence we are removing them.

:)


imath (approved):
I agree!

PR #408, opened 2021-06-22
Blogs: Remove user ID from avatar lookup

r-a-y:
Had a chance to use the `blogs` endpoint and noticed that the site avatar was always using the admin user's avatar instead of the site icon.

This PR fixes this by removing the `'user_id'` parameter from `bp_get_blog_avatar()`, which allows the proper site avatar to be used.


renatonascalves (approved):
Tks for that! I noticed that as well this past week. Had plans to update but seems I forgot all about it.

PR #410, opened 2021-07-03
Stop using the deprecated `bp_rest_get_user_url()` function

imath:
Now [r12985](https://buddypress.trac.wordpress.org/changeset/12985) has been committed to BuddyPress trunk, we need to now use `bp_rest_get_object_url()` instead.


renatonascalves (approved):
All good! Thanks! :)

PR #412, opened 2021-07-08
Improve the `BP_REST_Components_Endpoint` controller to inform about active component features

imath:
So far BP Blocks used WP Block Editor settings filters to get components active features. With the introduction of the Widget Block Editor, it appears it’s not a so great choice. Getting these features from blocks is a pain as they are not available from the same WP Core store. I believe we need to use our own BP Core Store with everything we need inside it. That’s what I’ve done into the latest version of the BP Blocks plugin. So far I was using a custom REST API fields but I think this should benefit everyone once we can inform from this endpoint about the acive BP features for each component.
I’ve also added a new `is_active` property to the schema because developers cannot rely on the `status` property as it is a l10n string. The `is_active` property is a boolean and is a better way to check whether a component is active or not.
The patch is using a brand new function I just added to BP Core see [r12987](https://buddypress.trac.wordpress.org/changeset/12987).

### DIFF
--- a/includes/bp-components/classes/class-bp-rest-components-endpoint.php
+++ b/includes/bp-components/classes/class-bp-rest-components-endpoint.php
@@ -167,7 +167,8 @@ class BP_REST_Components_Endpoint extends WP_REST_Controller {
 			)
 		);
 
-		if ( bp_current_user_can( 'manage_options' ) ) {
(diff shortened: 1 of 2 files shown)


renatonascalves (approved):
Pretty good!

I need to port this feature to the WP-CLI.

imath:
Thanks for your review and it's a very good idea to have it in the WP CLI Package 👍

PR #413, opened 2021-07-08
Add a new `registered_since` property to the Members schema

imath:
This is needed by the Dynamic Members Widget block when filtering members according to their registration date (newest filter).

See https://github.com/buddypress/bp-blocks/blob/8ac19a789b1bc94c3aaca2a9834e68f96477066d/src/bp-members/js/dynamic-members.js#L75 for context.


renatonascalves (approved):
Nice!

PR #414, opened 2021-07-09
Members: Only query for fields when necessary

r-a-y:
The WP REST API has a global `'_fields'` parameter allowing consumers of the API to restrict the response to only properties specified in `'_fields'`: https://developer.wordpress.org/rest-api/using-the-rest-api/global-parameters/#_fields

We should check this parameter before querying for any member properties in the `BP_REST_Members_Endpoint::user_data()` method: https://github.com/buddypress/BP-REST/blob/02d3de0ee278c52a80c2f77a272b9af1b079f4d6/includes/bp-members/classes/class-bp-rest-members-endpoint.php#L482

For example, if we perform the following BP Members API request: `example.com/wp-json/buddypress/v1/members/?_fields=id,name,user_login,link`, this shouldn't query other member fields like `'xprofile'`, `'friendship_status'`, etc. This can dramatically speed up the response time for the Members endpoint when the `'_fields'` global parameter is used.

From some local testing, this can speed up the Members endpoint by 50% or more. The main offender is the `'xprofile'` field. We should probably do the same  approach for our other endpoints as well where necessary, but there will not be as significant gains compared to the Members endpoint.

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -488,21 +488,47 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 			'id'                     => $user->ID,
 			'name'                   => $user->display_name,
 			'user_login'             => $user->user_login,
-			'link'                   => bp_core_get_user_domain( $user->ID, $user->user_nicename, $user->user_login ),
-			'member_types'           => bp_get_member_type( $user->ID, false ),
 			'roles'                  => array(),
 			'capabilities'           => array(),
 			'extra_capabilities'     => array(),
 			'registered_date'        => '',
-			'xprofile'               => $this->xprofile_data( $user->ID ),
 			'friendship_status'      => false,
 			'friendship_status_slug' => '',
 		);
 
+		// Get fields to output for REST endpoint.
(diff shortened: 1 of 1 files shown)


renatonascalves (approved):
I like this enhacement and feel the rest of the endpoints could benefit from it. I only don't like the implementation of it. But I'm not sure how core is doing it.

For this pull request, it is okay. :)

renatonascalves:
@r-a-y Could you fix this phpcs issue?

r-a-y:
> @r-a-y Could you fix this phpcs issue?

Done. I used `phpcbf --standard=WordPress` and that fixed the main short array syntax issue. It also added a bunch of spaces because WordPress loves spaces 😄

imath:
@r-a-y if you have some time, It would be nice to:
`git rebase -i master` to cherry pick commits that are only needed for this PR (after making sure your master branch is even with this repo one).

r-a-y:
> @r-a-y if you have some time, It would be nice to:
git rebase -i master to cherry pick commits that are only needed for this PR (after making sure your master branch is even with this repo one).

Thanks for the pointer about rebasing, imath 👍

I've just done this.

imath (approved):
Looks good! Thanks a lot @r-a-y 💪

PR #415, opened 2021-07-09
Add a new property `created_since` to the Groups item schema

imath:
This is needed by the Groups Dynamic Widget Block see [bp-blocks#52](https://github.com/buddypress/bp-blocks/pull/52/files#diff-28a91593413762c683a693126c2b9db73d07650c435d952f86136fea2f746e00R70)


renatonascalves (approved):
All good!

The failing unit test is not related to this and seems random. I'd disregard for now.

imath:
Thanks a lot for your review, I believe the same as you for the unit test 😉

PR #419, opened 2021-07-24
Remove activation key from response

renatonascalves:
Removing activation key in the CREATE response.


imath (approved):
looks good 👍  Thanks!

PR #421, opened 2021-08-14
Add GMT dates to component objects

### LINKED ISSUE #417: Add gmt date in activity, notifications and messages endpoints
is it possible to add the gmt date for every date field in activity, notifications, messages endpoints ?

an example on messages endpoint add date_gmt, start_date_gmt, date_sent_gmt etc...

renatonascalves:
fixes #417 

BuddyPress saves the dates in the database from its objects/components in GMT by default. The BP REST API has been advertising erroneously that the dates from the endpoints were respecting _the site's timezone_.

All dates were being passed to `bp_rest_prepare_date_response` which converts the date to the RFC3339 format, not that it is respecting the site's timezone.

This pull request adds two changes:

- First, we updated the default date field so that it uses `get_date_from_gmt`, which grabs the date  and returns it in the timezone of the site.
- Second, we are adding a new field with a `_gmt`, where the dates are in GMT (the same as the database).

Notes:

- `_gmt` dates are not embeddable. I'm not sure why but WordPress core does not seem to embed GMT dates. See an example here: https://developer.wordpress.org/rest-api/reference/posts/#schema
- Dates can return `null`, instead of `string` only;
- All dates were updated to `readonly => true`;
- Documentation will need to be updated.


imath (approved):
Thanks a lot for your work on this @renatonascalves It looks 👌

PR #423, opened 2021-08-27
Groups: when creating a group activity, save the item_id correctly

renatonascalves:
Issue raised at https://buddypress.trac.wordpress.org/ticket/8524

When creating a group activity, `groups_post_update` needs a `group_id` but the rest of the `create_item` and `update_item` actions need a `item_id`, which was missing.

### DIFF
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1040,13 +1040,13 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 		if ( ! empty( $schema['properties']['primary_item_id'] ) && ! empty( $request->get_param( 'primary_item_id' ) ) ) {
 			$item_id = (int) $request->get_param( 'primary_item_id' );
 
+			// Use a generic item ID.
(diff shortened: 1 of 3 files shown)


includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -1040,14 +1040,13 @@ protected function prepare_item_for_database( $request ) {
...
 			if ( bp_is_active( 'groups' ) && isset( $prepared_activity->component ) && buddypress()->groups->id === $prepared_activity->component ) {
 				$prepared_activity->group_id = $item_id;
-
-				// Use a generic item ID for other components.
-			} else {
-				$prepared_activity->item_id = $item_id;
 			}
+
+			// Use a generic item ID.
+			$prepared_activity->item_id = $item_id;
```
imath, line 1049: If I understand well, we need an `item_id` property by default, and a `group_id` property only when we create an activity attached to a group which uses the `groups_post_update()` function. If so I would put the code for the default value first, then the exception for the need of the `groups_post_update()` function.

imath (approved):
It looks good, I only have an "order" suggestion, but I leave you decide about the opportunity to apply it. Thanks a lot for your work on it.

Issue #425, opened 2021-10-16
Filter members by xprofile field


jakubrak:
Documentation states that result of querying list of members can be limited by xprofile field of string type:

https://developer.buddypress.org/bp-rest-api/reference/members/#arguments

however there is no information about string format. My suspcion is it has never worked, since BP_XProfile_Query expect an array, not a string: 
https://github.com/buddypress/BuddyPress-build/blob/master/bp-xprofile/classes/class-bp-xprofile-query.php#L59

I am right? If so, at least documentation should be clear that it is not implemented.

adiloztaser:
Hi @jakubrak,

I think you right. `xprofile` parameter type should be array. I've created a PR for this, let's wait @imath and @renatonascalves response.

PR #426, opened 2021-10-18
Update get_items xprofile parameter type

adiloztaser:
#425

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -126,7 +126,20 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 			$args['include'] = false;
 		}
 
-		if ( empty( $request->get_param( 'xprofile' ) ) ) {
+		if ( isset( $args['xprofile_query']['args'] ) && is_array( $args['xprofile_query']['args'] ) ) {
+			$xprofile_query_args = $args['xprofile_query']['args'];
+
+			if ( isset( $args['xprofile_query']['relation'] ) ) {
+				$xprofile_query_args = array_merge(
+					array(
+						'relation' => $args['xprofile_query']['relation'],
+					),
+					$xprofile_query_args
+				);
+			}
+
+			$args['xprofile_query'] = $xprofile_query_args;
+		} else {
 			$args['xprofile_query'] = false;
 		}
 
@@ -1139,8 +1152,7 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 
 		$params['xprofile'] = array(
 			'description'       => __( 'Limit results set to a certain XProfile field.', 'buddypress' ),
-			'default'           => array(),
-			'type'              => 'array',
+			'type'              => array( 'array', 'object' ),
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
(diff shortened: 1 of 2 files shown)


adiloztaser:
The test passed which is I've written for for this PR but I think failed test is not relevant with this PR.

imath:
You're welcome @adiloztaser, I might get some red now. In JavaScript you can't build associative arrays but objects, so to be usable we need to also allow the `xprofile` collection argument to be an object.  We also need to differentiate where to find the relation type and the list of queries. Here's the test I performed to check it's working:

```javascript
bp.apiRequest( {
	path: 'buddypress/v1/members',
	type: 'GET',
	data: {
		context: 'view',
		'xprofile': {
			relation: 'AND',
			args: [
				{
					field: 'Test field',
					value: 'Hello world',
					compare: '='
				}
			]
		},
	}
} ).done( function( data ) {
	data.forEach( function( member ) {
		console.log( member.name + ' => ' + member.id );
	} );
} ).fail( function( error ) {
	return error;
} );
```

Unfortunately xProfile queries do not seem to support fields with multiple values (the ones that generate checkboxes, multiselect ...).

adiloztaser:
Thank you for the improvement @imath, really good catch.

I've updated the test function after your changes because it wasn't using the `xprofile` parameter into the query. I think new test request more accurate and reliable.

tests/members/test-controller.php
```diff
@@ -218,6 +219,45 @@ public function test_get_items_with_types() {
...
+		$request->set_param( 'context', 'view' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 200, $response->get_status() );
+
+		$all_data = $response->get_data();
+		$this->assertNotEmpty( $all_data );
+
+		$this->assertTrue( 1 === count( $all_data ) );
+		$this->assertSame( array( $u ), wp_list_pluck( $all_data, 'id' ) );
```
renatonascalves, line 258: It would be good to test the second user is not in this array.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -126,7 +126,20 @@ public function get_items( $request ) {
...
 		}
 
-		if ( empty( $request->get_param( 'xprofile' ) ) ) {
+		if ( isset( $args['xprofile_query']['args'] ) && is_array( $args['xprofile_query']['args'] ) ) {
+			$xprofile_query_args = $args['xprofile_query']['args'];
+
+			if ( isset( $args['xprofile_query']['relation'] ) ) {
+				$xprofile_query_args = array_merge(
+					array(
+						'relation' => $args['xprofile_query']['relation'],
```
renatonascalves, line 135: Could you add a unit test using this relation?

renatonascalves (changes requested):
Looks good. I asked for two small changes before merging it.

tests/members/test-controller.php
adiloztaser: Thanks for advice. I've updated the test case for check this.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
adiloztaser: Good point, thanks for mentioning this. I've added new test cases both for `and` and `or` relations.

renatonascalves (approved):
Nice! This is good to go!

PR #430, opened 2021-11-14
XProfile Data: changing value type to a `string` instead of an `array`

### LINKED ISSUE #424: Can not update profile data for checkboxes and multiselectboxes.
I had the same problem a while ago, as mentioned in the below issue:
https://github.com/buddypress/BP-REST/issues/324

It was fixed back then, but now it is not working again. I don't know if some of the previous versions broke this.

Write now it is working with all fields data types, except checkboxes and multiselectboxes.

And it is not working with values with spaces or without spaces.

Thank you in advance.

renatonascalves:
fixes #424 

I'm suggesting we change the `value` type from an `array` into a `string`. The current behavior using an `array` is changing the output from `Option 01` into `["Option", "01"]`. Solving this for all use cases is tricky.

Instead of an `array` people would set the value as `Option 01,Option 02` which would be translated to an `array` in the REST API. This also would work for both single fields and fields with multiple options.

Other fixes:

- Unit tests support for the XProfile data update improved
- Clearing selected items of a field was introduced (a bug)
- Saving an option that a field does not have in its list of options returns an error

### DIFF
--- a/includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
+++ b/includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
@@ -70,12 +70,10 @@ class BP_REST_XProfile_Data_Endpoint extends WP_REST_Controller {
 					'permission_callback' => array( $this, 'update_item_permissions_check' ),
 					'args'                => array(
 						'value' => array(
-							'description' => __( 'The value(s) for the field data.', 'buddypress' ),
-							'required'    => true,
-							'type'        => 'array',
-							'items'       => array(
-								'type' => 'string',
-							),
+							'description'       => __( 'The value(s) for the field data.', 'buddypress' ),
+							'required'          => true,
+							'type'              => 'string',
+							'validate_callback' => 'rest_validate_request_arg',
 						),
 					),
 				),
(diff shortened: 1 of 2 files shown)


includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
```diff
@@ -70,12 +70,10 @@ public function register_routes() {
 					'permission_callback' => array( $this, 'update_item_permissions_check' ),
 					'args'                => array(
 						'value' => array(
-							'description' => __( 'The value(s) for the field data.', 'buddypress' ),
-							'required'    => true,
-							'type'        => 'array',
-							'items'       => array(
-								'type' => 'string',
-							),
+							'description'       => __( 'The value(s) for the field data.', 'buddypress' ),
```
imath, line 73: I'd probably specify the fact a comma separated list of values needs to be used in case of multiple values.

imath (approved):
Let's have it in! I just think we can improve the description of the parameter a bit 😉

includes/bp-xprofile/classes/class-bp-rest-xprofile-data-endpoint.php
renatonascalves: Updated as suggested!

Issue #433, opened 2021-11-25
List user groups including hidden groups is not working for non administrator


marioshtika:
Hello,

I want to get all the groups and I am using the `GET /buddypress/v1/groups` endpoint as shown in the [documentation](https://developer.buddypress.org/bp-rest-api/reference/user-groups/groups/#list-user-groups)

I want to include hidden groups too, so I am adding the `show_hidden=true` parameter.

```
bp.apiRequest( {
  path: 'buddypress/v1/groups',
  type: 'GET',
  data: {
    context: 'view',
    show_hidden: true
  }
} ).done( function( data ) {
  return data;
} ).fail( function( error ) {
  return error;
} );
```

This is working fine if I am the creator of the hidden group.

If the user making the request is a member of the hidden group (and not the creator) the hidden group will not show in the response.

Any idea, why this is happening?

Thank you in advance.

renatonascalves:
This is happening because you are not setting the `user_id` param. This param is required when you are trying to get groups from a specific member. Internally, we check if the `user_id` is the same as the current logged in user. If they are, we include hidden or private groups.

This is necessary because this endpoint can return public groups even for non-logged in users.

marioshtika:
Hello @renatonascalves and thank you for your answer.

I will try to explain my scenario and let me know if there is something I am doing wrong or if this is not the proper behavior.

If I use the `user_id` parameter it will show me the groups I am a member of.

What I want to achieve is getting all groups available to me and if I want to use `show_hidden: true` I would expect as an addition to my available groups to see the hidden groups I am a member of.

Here is an example of my scenario

I have four groups
- Group 1 (public) - I am a member of this group
- Group 2 (public) - I am **not** a member of this group
- Group 3 (private) - I am a member of this group
- Group 4 (hidden) - I am a member of this group

If I retrieve all groups, without `show_hidden: true` it will show me 
- Group 1
- Group 2
- Group 3

If I retrieve all groups, with the `show_hidden: true` it will show me again 
- Group 1
- Group 2
- Group 3

_It will not show me Group 4_

If I retrieve all groups, with the `show_hidden: true` and `user_id: ___` it will show me 
- Group 1
- Group 3
- Group 4

_It will not show me Group 2 which is public_

Is there anything we can do for this scenario?

Thank you in advance

renatonascalves:
If you are returning all groups you are a member (when you use the `user_id` parameter, that's what you are doing), you won't have public groups where **you are not** a member that group.

[...]

PR #434, opened 2021-12-05
Groups: improving unit tests

renatonascalves:
Improving groups unit tests, particularly related to the endpoint that get groups.

Related to #433


imath (approved):
Looks good, thanks for these improvements 👍

PR #447, opened 2022-07-20
Add action links to items

### LINKED ISSUE #444: Add links for item available actions
I'm currently working on adding a BP reNouveau template pack in BP 11.0.0. This template pack will only use the BP REST API. So I'll probably report other possible improvements.

This one is about being able to get the endpoint for action concerning a member item of a members loop (eg: the endpoints to add friend/cancel friendship)

I see the WP Posts REST Controller is adding "self" links to add available actions, see https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1939|L1947

It would be very useful to have this kind of functionality for members actions, but also for other component items (activity, group..) actions.

renatonascalves:
fixes #444

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -66,11 +66,7 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 					'callback'            => array( $this, 'get_item' ),
 					'permission_callback' => array( $this, 'get_item_permissions_check' ),
 					'args'                => array(
-						'context'         => $this->get_context_param(
-							array(
-								'default' => 'view',
-							)
-						),
+						'context'         => $this->get_context_param( array( 'default' => 'view' ) ),
 						'populate_extras' => array(
 							'description'       => __( 'Whether to fetch extra BP data about the returned group.', 'buddypress' ),
 							'context'           => array( 'view', 'edit' ),
@@ -951,9 +947,7 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 	 * @return array
 	 */
 	protected function prepare_links( $group ) {
-		$base = sprintf( '/%s/%s/', $this->namespace, $this->rest_base );
-
-		// Entity meta.
+		$base  = sprintf( '/%s/%s/', $this->namespace, $this->rest_base );
 		$links = array(
 			'self'       => array(
 				'href' => rest_url( $base . $group->id ),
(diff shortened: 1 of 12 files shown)


renatonascalves:
@imath Could you take a look at the actions and confirm if I'm missing any?

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -737,7 +737,7 @@ public function prepare_item_for_response( $item, $request ) {
 			'types'              => bp_groups_get_group_type( $item->id, false ),
 			'admins'             => array(),
 			'mods'               => array(),
-			'total_member_count' => null,
+			'total_member_count' => 0,
 			'last_activity'      => null,
 			'last_activity_diff' => null,
 		);
```
imath, line 743: When we list groups into the Component's directory, and if the [user is not banned from the group](https://github.com/buddypress/buddypress/blob/master/src/bp-groups/bp-groups-template.php#L3583), we include a button to:
- request a membership for a private group (if not already a member of it)
- join a public group (if not already a member of it)
- leave a group (if already a member of it)
- accept an invitation to join a private group

See https://github.com/buddypress/buddypress/blob/master/src/bp-groups/bp-groups-template.php#L3574

I believe we should include action links dynamically according to the existing relation with the group.
- The logged in user is invited to join the group, there should be 2 action links: **accept** and **reject** invite.
- The logged in user is not a member of a public group, there should be 1 action link: **join**.
- The logged in user is a member of a group, there should be 1 action link: **leave**.
- The logged in user is not a member of a private group, there should be 1 action link: **request a membership**.
- The logged in user requested a membership to a private group, there should be 1 action link: **withdraw the request**.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -479,6 +482,59 @@ public function prepare_item_for_response( $user, $request ) {
...
+				$friends_component = buddypress()->friends->id;
+
+				$links['bp-action-create-friendship'] = array(
+					'href' => rest_url( sprintf( '/%1$s/%2$s/', $this->namespace, $friends_component ) ),
+				);
+
+				$links['bp-action-manage-friendship'] = array(
+					'href' => bp_rest_get_object_url( $user->ID, $friends_component ),
+				);
+			}
```
imath, line 517: Here again these action links should be more dynamic according to the friendship relation of the logged in user with the listed user. See https://github.com/buddypress/buddypress/blob/master/src/bp-friends/bp-friends-template.php#L373

I believe we should use the existing `friendship_status_slug` schema property to decide what action links should be included, eg:
- `not_friend` -> an action link to init a friendship (add friend)
- `is_friend` -> an action link to delete the friendship
- `pending` -> an action link to cancel the friendship request
- `awaiting_response` -> 2 action links: one to accept & another one to reject the friendship request.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -479,6 +482,59 @@ public function prepare_item_for_response( $user, $request ) {
...
+
+			if ( bp_is_active( 'messages' ) ) {
```
imath, line 519: Private message are restricted to friends by default unless `buddypress()->messages->autocomplete_all` is `true`

imath (changes requested):
Thanks for your work on this. I believe we should build these action links in a more dynamic way: only include the available actions according to the existing relationship between the logged in user and the object.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
renatonascalves: Added as suggested!

includes/bp-members/classes/class-bp-rest-members-endpoint.php
renatonascalves: Added as suggested!

renatonascalves:
@imath Could you take another look? :)

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Looks great, thanks you 👍

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -487,36 +487,57 @@ public function prepare_item_for_response( $user, $request ) {
...
+						$links['bp-action-reject-friendship-request'] = $friends_action;
+						break;
+				}
 			}
 
-			if ( bp_is_active( 'messages' ) ) {
+			if (
+				bp_is_active( 'messages' )
+				&& true === wp_validate_boolean( $user_data['friendship_status'] )
+				&& buddypress()->messages->autocomplete_all === false
```
imath, line 539: I believe it should be `&& ( true === wp_validate_boolean( $user_data['friendship_status'] ) || buddypress()->messages->autocomplete_all === true )`

It's possible to send PM when members are friends or autocomplete all is true.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -979,13 +973,74 @@ protected function prepare_links( $group ) {
...
 		}
 
+		// Actions.
+
+		$user_id = bp_loggedin_user_id();
+
+		if ( is_user_logged_in() && false === bp_group_is_user_banned( $group->id, $user_id ) ) {
+			$membership_namespace = "{$this->rest_base}/membership-requests";
+			$request_id           = groups_check_for_membership_request( $user_id, $group->id );
+			$is_group_member      = $group->is_member;
```
imath, line 983: `$group->is_member` is an int, we need to cast it as boolean: `(bool) $group->is_member;`

imath (dismissed):
Thanks a lot for your improvements. It's fine for friendship, there are just tiny fixes to bring and it should be good to be merged?

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
imath: Thanks 👍

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
imath: Oh and I forgot, I was a bit surprised by the `switch( true )` thing. Could we use an `if {} elseif {}...` instead ?

includes/bp-members/classes/class-bp-rest-members-endpoint.php
renatonascalves: Good catch! I misunderstood theh autocomplete part.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
renatonascalves: It's also a good approach to avoid complex if/else statements. Like the scenario here.

imath: ok, fine about keeping the `switch( true )` approach. I'll try to get used to it :)

renatonascalves: By the way, this is actually why I assumed it was a `bool`: https://github.com/buddypress/buddypress/blob/5cfd97515673dacc477902292547e268f19d1fa3/src/bp-groups/classes/class-bp-groups-group.php#L126

imath (approved):
Ready to go! Thanks again 👏

PR #448, opened 2022-07-20
Activity: Improvements to the `update_item`

### LINKED ISSUE #446: Activity `update_item()` method needs improvements
I'm also working on the Activity Block Editor and noticed 2 possible improvements: 

1. the date is not kept although I've sent it. This can be annoying if an Admin edits the activity of another user. In other words in the edited activated date was 'before' using `update_item()` makes it become 'now'. I believe it's in the method to prepare the item for the database the fix needs to happen.
2. When updating an activity about a group, `groups_record_activity()` should be used to make sure activity about non public groups have their visibility unchanged. In one of my tests, when I've updated a public "my profile" activity to attach it to a private group, the visibility of this activity wasn't hidden but public.

Thanks in advance for your help 😄

renatonascalves:
fixes #446


imath (approved):
Perfect for point 1. This is what I wanted. Thanks a lot and sorry if my explanations were a bit confusing.

About point 2, what do you think of using `groups_record_activity()` when updating a group activity? The downside of `bp_activity_add()` is that you need to resend the `hide_sitewide` prop :(

renatonascalves:
@imath About point 1: it is clear that even if a user, the creator of the activity, updates his own activity, the date won't be updated. This will affect not only updates from admins.

2: Let me review the code of this feature. I'm not following 100% the suggested approach.

imath:
@renatonascalves I've just pushed a commit that should help you understand the problem I was talking about 😃 . I realized the `hidden` property of the Activity item schema was never set, so I've fixed this. Finally instead of using `groups_record_activity()` I believe we can carry on using `bp_activity_add()` as soon as we make sure Private and Hidden group activities always remain `hidden => true`. Otherwise it would make these private objects activities visible for everyone.

renatonascalves:
Understood!

I just merged `master` locally and there seems to be a merge conflict. Could you activate the branch settings to always update the pr with the main branch? So that they are in sync.

imath:
Hum this is weird as I've rebased this branch with master before editing it and I see I can safely merge it to master from GH... I'm not sure what setting you want me to activate, here's a screenshot:
Capture d’écran 2022-07-30 à 09 43 00

Can you guide me?

renatonascalves:
@imath  Can you see this in the branch protection rules?

Screen Shot 2022-07-30 at 12 31 11

imath:
@renatonascalves I just added a protection rule to master branch checking that option 👍

renatonascalves:
Thank youuuu! 🥳

PR #450, opened 2022-07-30
Components: check if the `core` component is in the `array`

renatonascalves:
As titled.

Bug reported here: https://buddypress.org/support/topic/fatal-error-after-updating-to-php-8-0/

### DIFF
--- a/includes/bp-components/classes/class-bp-rest-components-endpoint.php
+++ b/includes/bp-components/classes/class-bp-rest-components-endpoint.php
@@ -99,7 +99,7 @@ class BP_REST_Components_Endpoint extends WP_REST_Controller {
 
 		// Core component is always active.
 		if ( 'optional' !== $type && ! empty( $components['core'] ) ) {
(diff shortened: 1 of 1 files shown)


imath (approved):
Looks good! Thanks a lot for the fix 👍

includes/bp-components/classes/class-bp-rest-components-endpoint.php
```diff
@@ -95,11 +95,11 @@ public function get_items( $request ) {
 		$components = bp_core_get_components( $type );
 
 		// Active components.
-		$active_components = apply_filters( 'bp_active_components', bp_get_option( 'bp-active-components' ) );
+		$active_components = (array) apply_filters( 'bp_active_components', bp_get_option( 'bp-active-components' ) );
```
renatonascalves, line 98: The issue was here. This option might return a `string`. We need to cast it as an `integer` to avoid `Cannot access offset of type string on string`.

PR #452, opened 2022-08-07
PHPCS: Using BP ruleset

renatonascalves:
As titled!


imath:
@renatonascalves here's the reason tests are failing with WP Trunk > https://buddypress.trac.wordpress.org/ticket/8727

PR #454, opened 2022-08-14
Upgrade plugin with latest changes

renatonascalves:
- Run GH action on the `master` branch.
- Bump some GH actions packages
- Test against PHP 8.1
- Let `yoast/phpunit-polyfills` decide which PHPUnit version to install based on the current PHP version


renatonascalves:
@imath I don't have permission to do that, but could update the branch protection rules to allow merging on `master` only if the GH actions passes?

imath:
@renatonascalves sure but can you guide me on doing so as I don't see how to make it possible 😬. It looks like we need to use the status checks thing but it's unclear how to set this up!
Capture d’écran 2022-08-15 à 11 42 24
When I search for status checks of the last week I get no results 😞

renatonascalves:
Weird. So when you search for WP: 6.0 or PHPCS you get no actions?

imath:
Thanks for your help, here's what I get, should I select all ?
Capture d’écran 2022-08-15 à 14 13 16

renatonascalves:
IMO, all but 8.1.

imath:
Ok, just did it!

imath (approved):
Looks good! Thanks for your work on it 👍

PR #459, opened 2022-10-02
Add the `embed` context to the `name`’s members endpoint schema property

### LINKED ISSUE #458: activity display name
Hi 
can you please add the activity creator display name to the activities list returned, i know we have the user id, but making a call  for each activity to get the user display name its really not good idea

 thank you

image4

imath:
To get the Activity Author's display name, when making an `activity/?_embed=true` request, we need to add the `embed` context to the `name` property of the members endpoint schema.

Fixes #458


renatonascalves (approved):
Good catch! 💯

imath:
Thanks for approving the PR @renatonascalves 😍

PR #460, opened 2022-10-03
Simplify the group membership check into the Activity controller

imath:
Group Admin or Group Mod is firstly a group member, we just need to check whether a member is a member of a private/hidden group to allow hidden activities to be fetched.

I also think we should provide a filter to eventually allow third party component to manage hidden activities.

I've been looking at this part because I got an error having the `groups` component inactive and trying to get activity comments using `activity` as the component attribute. Problem was, when the `groups` component is not active, `buddypress()->groups->id is also not set.


renatonascalves (approved):
looks good!

I have vague memories. I think I added that way to avoid a group member from seeing someone else private activities. But I might be mistaken.

PR #463, opened 2022-10-27
Thread: update recipients `array`

renatonascalves:
Make sure we return a proper `array` so that the output is consistent with the rest of the endpoints.


imath (approved):
Good catch, you're right this was my mistake, it wasn't on purpose. Thanks a lot for fixing it 💪👏

PR #464, opened 2022-10-28
Thread: add name to the recipient response

renatonascalves:
See: https://buddypress.trac.wordpress.org/ticket/8698


imath:
This looks good 👍. Do you have an idea of what's going wrong with unit tests?

imath (approved):
Thanks for your work on it 👍

PR #467, opened 2022-11-24
Implement the `BP_REST_Members_Endpoint::delete_current_item()` method

imath:
This is a follow up of https://buddypress.trac.wordpress.org/ticket/8758

After more research, we haven't implemented this method yet. WordPress is kind of "faking" this method as it requires the user to have the `delete_users` capability which regular users don't. As we allow users to delete their account from the front-end, we should also allow users to delete their account using the BP REST API.

Ideally, I'd like this to be committed before we release BuddyPress 11.0.0 (~ December 14)

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -415,13 +415,27 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 	 * @return true|WP_Error True if the request has access to delete the item, WP_Error object otherwise.
 	 */
 	public function delete_current_item_permissions_check( $request ) {
-		$request['id'] = get_current_user_id();
+		$retval = new WP_Error(
+			'bp_rest_authorization_required',
+			__( 'Sorry, you are not allowed delete your account.', 'buddypress' ),
+			array(
+				'status' => rest_authorization_required_code(),
+			)
+		);
 
-		if ( bp_disable_account_deletion() ) {
-			return parent::delete_item_permissions_check( $request );
+		if ( ! bp_disable_account_deletion() ) {
+			$retval = true;
 		}
 
-		return true;
+		/**
+		 * Filter the members `delete_current_item` permissions check.
+		 *
+		 * @since 0.7.0
+		 *
+		 * @param true|WP_Error   $retval  Returned value.
+		 * @param WP_REST_Request $request The request sent to the API.
+		 */
+		return apply_filters( 'bp_rest_members_delete_current_item_permissions_check', $retval, $request );
 	}
 
 	/**
@@ -439,7 +453,7 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 		$user = bp_rest_get_user( get_current_user_id() );
 
 		if ( ! $user instanceof WP_User ) {
-			$retval = new WP_Error(
+			return new WP_Error(
 				'bp_rest_member_invalid_id',
 				__( 'Invalid member ID.', 'buddypress' ),
 				array(
--- a/includes/bp-members/classes/class-bp-rest-members-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-members-endpoint.php
@@ -406,45 +406,71 @@ class BP_REST_Members_Endpoint extends WP_REST_Users_Controller {
 	}
 
 	/**
-	 * Deleting the current user is not implemented into this endpoint.
-	 *
-	 * This action is specific to the User Settings endpoint.
+	 * Checks if a given request has access to delete the current user.
 	 *
 	 * @since 0.1.0
+	 * @since 0.7.0 Do implement this method.
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return WP_Error                WP_Error object to inform it's not implemented.
+	 * @return true|WP_Error True if the request has access to delete the item, WP_Error object otherwise.
 	 */
 	public function delete_current_item_permissions_check( $request ) {
-		return new WP_Error(
-			'bp_rest_invalid_method',
-			/* translators: %s: transport method name */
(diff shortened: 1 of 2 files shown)


includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -406,45 +406,71 @@ public function delete_item_permissions_check( $request ) {
...
-				'status' => 405,
-			)
-		);
+		$request['id'] = get_current_user_id();
+
+		if ( bp_disable_account_deletion() ) {
+			return parent::delete_item_permissions_check( $request );
+		}
+
+		return true;
```
renatonascalves, line 424: Do we need to add a hook as we do for other permissions check?

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -406,45 +406,71 @@ public function delete_item_permissions_check( $request ) {
...
-		return new WP_Error(
-			'bp_rest_invalid_method',
-			/* translators: %s: transport method name */
-			sprintf( __( '\'%s\' Transport method not implemented.', 'buddypress' ), $request->get_method() ),
+		$request->set_param( 'context', 'edit' );
+
+		$user = bp_rest_get_user( get_current_user_id() );
+
+		if ( ! $user instanceof WP_User ) {
+			$retval = new WP_Error(
```
renatonascalves, line 442: I think you meant to return this error instead of assigning it to this variable.

```suggestion
			return new WP_Error(
```

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -406,45 +406,71 @@ public function delete_item_permissions_check( $request ) {
...
+				'bp_rest_member_invalid_id',
+				__( 'Invalid member ID.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+		}
+
+		$previous = $this->prepare_item_for_response( $user, $request );
+		$result   = bp_core_delete_account();
```
renatonascalves, line 452: What happens to the current user after they delete themselves? Do we need to clear any source of nonce programmatically?

renatonascalves (approved):
Good catch!

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Absolutely, thanks for the reminder 👍

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Ouch, right! Thanks a lot!

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Once the user is deleted, every subsequent request will return a `rest_cookie_invalid_nonce` error until the page is refreshed. I think  it's fine as when you delete your account, your next action is to leave the site 😉

renatonascalves: Cool! Just to be clear, via the REST API, after deleting your account, any REST API request will return an error, right?!

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -406,45 +406,85 @@ public function delete_item_permissions_check( $request ) {
...
 		);
+
+		if ( ! bp_disable_account_deletion() ) {
+			$retval = true;
+		}
+
+		/**
+		 * Filter the members `delete_current_item` permissions check.
+		 *
+		 * @since 0.7.0
```
renatonascalves, line 433: For documentation purposes, this is weird. Since it was introduced in the BP 11 version but it will show in the developer.bp site was 0.7.0.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Absolutely

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: I agree, but we've been doing so for a while. I think as long as we are cautious making sure to respect version links between BuddyPress and the BP REST API, it's fine.

- BP REST API 0.1.0 is  linked to BP 5.0.0
- BP REST API 0.2.0 is  linked to BP 6.0.0
- BP REST API 0.3.0 is  linked to BP 7.0.0
- BP REST API 0.4.0 is  linked to BP 8.0.0
- BP REST API 0.5.0 is  linked to BP 9.0.0
- BP REST API 0.6.0 is  linked to BP 10.0.0
- BP REST API 0.7.0 is  linked to BP 11.0.0

For 0.1.0 I was replacing 0.1.0 by 5.0.0 inside a Grunt task. Maybe we can improve this task:
https://github.com/buddypress/buddypress/blob/master/Gruntfile.js#L294|L305

renatonascalves: Should we replace then, in the future, so that the phpDoc documentation reflects the right version? Or is it the plan to document that somehow as well?

I don't mind the difference, I'm just concerned this might cause confusion for those unaware of why they are different.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
renatonascalves: Awesome! Thanks for confirming! :)

includes/bp-members/classes/class-bp-rest-members-endpoint.php
imath: Replacing them should ease the process I agree, let's do it soon!

renatonascalves: Cool! Tks! :)

PR #469, opened 2022-12-11
Adds a new collection parameter to include a list of profile fields groups

imath:
The `profile_group_id` parameter of `BP_XProfile_Group::get()` now accepts (BuddyPress 11.0.0) an array of profile group ids. Instead of modifying the `profile_group_id` parameter of the collection params, I suggest to introduce an `include_groups` one.

See https://buddypress.trac.wordpress.org/changeset/13358/

### DIFF
--- a/includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
+++ b/includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
@@ -166,9 +166,9 @@ class BP_REST_XProfile_Field_Groups_Endpoint extends WP_REST_Controller {
 			$args['exclude_groups'] = false;
 		}
 
-		$include_groups = $request->get_param( 'include_groups' );
-		if ( $include_groups && ! $args['profile_group_id'] ) {
-			$args['profile_group_id'] = $include_groups;
+		$include = $request->get_param( 'include' );
+		if ( $include && ! $args['profile_group_id'] ) {
+			$args['profile_group_id'] = $include;
 		}
 
 		/**
@@ -866,12 +866,12 @@ class BP_REST_XProfile_Field_Groups_Endpoint extends WP_REST_Controller {
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
-		$params['include_groups'] = array(
+		$params['include'] = array(
 			'description'       => __( 'Ensure result set inludes specific profile field groups.', 'buddypress' ),
 			'default'           => array(),
 			'type'              => 'array',
 			'items'             => array( 'type' => 'integer' ),
-			'sanitize_callback' => 'bp_rest_sanitize_string_list',
+			'sanitize_callback' => 'wp_parse_id_list',
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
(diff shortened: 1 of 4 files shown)


includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
```diff
@@ -861,6 +866,15 @@ public function get_collection_params() {
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
+		$params['include_groups'] = array(
```
renatonascalves, line 869: `include` would be more in line with other endpoints. And since this is a field group endpoint, that might already be implied.

```suggestion
		$params['include'] = array(
```

includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
```diff
@@ -861,6 +866,15 @@ public function get_collection_params() {
...
+			'description'       => __( 'Ensure result set inludes specific profile field groups.', 'buddypress' ),
+			'default'           => array(),
+			'type'              => 'array',
+			'items'             => array( 'type' => 'integer' ),
+			'sanitize_callback' => 'bp_rest_sanitize_string_list',
```
renatonascalves, line 874: Since this is an array of ids...

```suggestion
			'sanitize_callback' => 'wp_parse_id_list',
```

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -1169,6 +1174,15 @@ public function get_collection_params() {
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
+		$params['include_groups'] = array(
+			'description'       => __( 'Ensure result set inludes specific profile field groups.', 'buddypress' ),
+			'default'           => array(),
+			'type'              => 'array',
+			'items'             => array( 'type' => 'integer' ),
+			'sanitize_callback' => 'bp_rest_sanitize_string_list',
```
renatonascalves, line 1182:
```suggestion
			'sanitize_callback' => 'wp_parse_id_list',
```

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -1169,6 +1174,15 @@ public function get_collection_params() {
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
+		$params['include_groups'] = array(
```
renatonascalves, line 1177: Here it makes sense `include_group` since this is a fields endpoint.

tests/testcases/xprofile/test-group-controller.php
```diff
@@ -63,6 +63,37 @@ public function test_get_items() {
 		}
 	}
 
+	/**
+	 * @group get_items
+	 */
+	public function test_get_items_include_groups() {
```
renatonascalves, line 69: it would be good to have a unit test for the field controller.

renatonascalves (approved):
Looks good! I made a few suggestions, though.

includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
imath: A bit below, there's the `exclude_groups` parameter, that's why I've used `include_groups` here. Just `include` is fine with me. I'll use this. Thanks for your suggestion.

renatonascalves: I see! In that case, `include_groups` would be a better option to keep it consistent.

PR #473, opened 2023-02-01
Update wp-env to latest version

imath:
This is also updating simple git to 3.16.

Fixes #472


imath:
Looks like it needs more work about how WP PHPUnit constants are defined. @renatonascalves do you have an idea about what's going wrong?

PR #487, opened 2023-07-25
Support the bp_view capability.

dcavins:
This new capability will work with the community visibility features coming in BuddyPress 12.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/8734

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
@@ -169,17 +169,27 @@ class BP_REST_Attachments_Group_Avatar_Endpoint extends WP_REST_Controller {
 	 * @return true|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-		$retval      = new WP_Error(
-			'bp_rest_group_invalid_id',
-			__( 'Invalid group ID.', 'buddypress' ),
-			array(
-				'status' => 404,
-			)
-		);
-		$this->group = $this->groups_endpoint->get_group_object( $request );
+		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+			$retval = new WP_Error(
+				'bp_rest_authorization_required',
+				__( 'Sorry, you cannot view group details.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		} else {
+			$retval      = new WP_Error(
+				'bp_rest_group_invalid_id',
+				__( 'Invalid group ID.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+			$this->group = $this->groups_endpoint->get_group_object( $request );
 
-		if ( false !== $this->group ) {
-			$retval = true;
+			if ( false !== $this->group ) {
+				$retval = true;
+			}
 		}
 
 		/**
--- a/includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
@@ -169,15 +169,15 @@ class BP_REST_Attachments_Group_Avatar_Endpoint extends WP_REST_Controller {
 	 * @return true|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you cannot view group details.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		} else {
+		$retval = new WP_Error(
+			'bp_rest_authorization_required',
+			__( 'Sorry, you cannot view group details.', 'buddypress' ),
+			array(
+				'status' => rest_authorization_required_code(),
+			)
+		);
+
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
 			$retval      = new WP_Error(
 				'bp_rest_group_invalid_id',
 				__( 'Invalid group ID.', 'buddypress' ),
--- a/includes/bp-groups/classes/class-bp-rest-attachments-group-cover-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-attachments-group-cover-endpoint.php
@@ -155,17 +155,27 @@ class BP_REST_Attachments_Group_Cover_Endpoint extends WP_REST_Controller {
 	 * @return true|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-		$retval      = new WP_Error(
-			'bp_rest_group_invalid_id',
-			__( 'Invalid group id.', 'buddypress' ),
-			array(
-				'status' => 404,
-			)
-		);
-		$this->group = $this->groups_endpoint->get_group_object( $request );
+		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+			$retval = new WP_Error(
+				'bp_rest_authorization_required',
+				__( 'Sorry, you cannot view group details.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
+				)
+			);
+		} else {
+			$retval      = new WP_Error(
+				'bp_rest_group_invalid_id',
+				__( 'Invalid group ID.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
+			$this->group = $this->groups_endpoint->get_group_object( $request );
 
-		if ( false !== $this->group ) {
-			$retval = true;
+			if ( false !== $this->group ) {
+				$retval = true;
+			}
 		}
 
 		/**
--- a/includes/bp-groups/classes/class-bp-rest-attachments-group-cover-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-attachments-group-cover-endpoint.php
@@ -155,15 +155,15 @@ class BP_REST_Attachments_Group_Cover_Endpoint extends WP_REST_Controller {
 	 * @return true|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
-		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
-			$retval = new WP_Error(
-				'bp_rest_authorization_required',
-				__( 'Sorry, you cannot view group details.', 'buddypress' ),
-				array(
-					'status' => rest_authorization_required_code(),
-				)
-			);
-		} else {
+		$retval = new WP_Error(
+			'bp_rest_authorization_required',
+			__( 'Sorry, you cannot view group details.', 'buddypress' ),
+			array(
+				'status' => rest_authorization_required_code(),
+			)
+		);
+
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
 			$retval      = new WP_Error(
 				'bp_rest_group_invalid_id',
 				__( 'Invalid group ID.', 'buddypress' ),
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -199,7 +199,7 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 		);
 
 		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
-			$group  = $this->groups_endpoint->get_group_object( $request->get_param( 'group_id' ) );
+			$group = $this->groups_endpoint->get_group_object( $request->get_param( 'group_id' ) );
 			if ( empty( $group->id ) ) {
 				$retval = new WP_Error(
 					'bp_rest_group_invalid_id',
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -197,18 +197,20 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 				'status' => rest_authorization_required_code(),
 			)
 		);
-		$group  = $this->groups_endpoint->get_group_object( $request->get_param( 'group_id' ) );
 
-		if ( empty( $group->id ) ) {
-			$retval = new WP_Error(
-				'bp_rest_group_invalid_id',
-				__( 'Invalid group ID.', 'buddypress' ),
-				array(
-					'status' => 404,
-				)
-			);
-		} elseif ( bp_current_user_can( 'bp_moderate' ) || 'public' === $group->status || groups_is_user_member( bp_loggedin_user_id(), $group->id ) ) {
-			$retval = true;
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+			$group  = $this->groups_endpoint->get_group_object( $request->get_param( 'group_id' ) );
+			if ( empty( $group->id ) ) {
+				$retval = new WP_Error(
+					'bp_rest_group_invalid_id',
+					__( 'Invalid group ID.', 'buddypress' ),
+					array(
+						'status' => 404,
+					)
+				);
+			} elseif ( bp_current_user_can( 'bp_moderate' ) || 'public' === $group->status || groups_is_user_member( bp_loggedin_user_id(), $group->id ) ) {
+				$retval = true;
+			}
 		}
 
 		/**
--- a/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
@@ -204,19 +204,20 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return true|WP_Error
+	 * @return boolean
 	 */
 	public function get_items_permissions_check( $request ) {
+		$retval = bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) );
 
 		/**
 		 * Filter the groups `get_items` permissions check.
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param true $value True.
+		 * @param boolean         $retval  Whether the user has access to groups component items.
 		 * @param WP_REST_Request $request The request sent to the API.
 		 */
-		return apply_filters( 'bp_rest_groups_get_items_permissions_check', true, $request );
+		return apply_filters( 'bp_rest_groups_get_items_permissions_check', $retval, $request );
 	}
 
 	/**
@@ -268,18 +269,21 @@ class BP_REST_Groups_Endpoint extends WP_REST_Controller {
 				'status' => rest_authorization_required_code(),
 			)
 		);
-		$group  = $this->get_group_object( $request );
 
-		if ( empty( $group->id ) ) {
-			$retval = new WP_Error(
-				'bp_rest_group_invalid_id',
-				__( 'Invalid group ID.', 'buddypress' ),
-				array(
-					'status' => 404,
-				)
-			);
-		} elseif ( $this->can_see( $group ) ) {
-			$retval = true;
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+			$group  = $this->get_group_object( $request );
+
+			if ( empty( $group->id ) ) {
+				$retval = new WP_Error(
+					'bp_rest_group_invalid_id',
+					__( 'Invalid group ID.', 'buddypress' ),
+					array(
+						'status' => 404,
+					)
+				);
+			} elseif ( $this->can_see( $group ) ) {
+				$retval = true;
+			}
 		}
 
 		/**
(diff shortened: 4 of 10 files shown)


imath:
Hi @dcavins 

Thanks a lot for your patch, I saw only Members/Groups and xProfile endpoints were updated, shouldn't we do the same for the other component endpoints (except perhaps the signup one)?

dcavins:
Hi @imath,

Yes, I did not add `bp_view` checks in permission checks that were already more strict (like they already required that the user be logged in or have the bp_moderate capability). For instance:
https://github.com/buddypress/BP-REST/blob/master/includes/bp-friends/classes/class-bp-rest-friends-endpoint.php#L174

However, I can imagine adding the check so that plugins, etc, have a way to filter access to those items, even though in core, they probably won't be adding a stricter check, just potentially duplicating a check. What do you think?

imath:
Ah right 😅. I believe there are already filters in place into endpoints, so it's probably not an emergency for now, ideally it would probably be best to improve the way we check caps: but it can wait a bit longer. Thanks for your feedback.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
```diff
@@ -169,17 +169,27 @@ public function get_item( $request ) {
...
 	public function get_item_permissions_check( $request ) {
-		$retval      = new WP_Error(
-			'bp_rest_group_invalid_id',
-			__( 'Invalid group ID.', 'buddypress' ),
-			array(
-				'status' => 404,
-			)
-		);
-		$this->group = $this->groups_endpoint->get_group_object( $request );
+		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
```
renatonascalves, line 172: Regardless if you have this cap or not, we should check if the group ID exists **first**. Then we check for the user's permissions to view the group. 

Imagine a mobile app handling this response: should the app show a 404 page/screen or redirect the user to the login screen?

The error here might pass the wrong status of the _document_.

includes/bp-groups/classes/class-bp-rest-attachments-group-cover-endpoint.php
```diff
@@ -155,17 +155,27 @@ public function get_item( $request ) {
...
 	public function get_item_permissions_check( $request ) {
-		$retval      = new WP_Error(
-			'bp_rest_group_invalid_id',
-			__( 'Invalid group id.', 'buddypress' ),
-			array(
-				'status' => 404,
-			)
-		);
-		$this->group = $this->groups_endpoint->get_group_object( $request );
+		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
```
renatonascalves, line 158: The same suggestion I left for the group avatar.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
```diff
@@ -169,17 +169,27 @@ public function get_item( $request ) {
...
+			$retval = new WP_Error(
+				'bp_rest_authorization_required',
+				__( 'Sorry, you cannot view group details.', 'buddypress' ),
+				array(
+					'status' => rest_authorization_required_code(),
```
renatonascalves, line 177: I'm also not sure this is the right error for this feature.

This function returns 401 if the user is not logged in, 403 if the user is logged in. I could be logged in and based on this feature flag, get a 401. A incorrect error code.

renatonascalves: We might need to investigate the recommended error code for this situation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -197,18 +197,20 @@ public function get_items_permissions_check( $request ) {
...
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+			$group  = $this->groups_endpoint->get_group_object( $request->get_param( 'group_id' ) );
+			if ( empty( $group->id ) ) {
+				$retval = new WP_Error(
+					'bp_rest_group_invalid_id',
+					__( 'Invalid group ID.', 'buddypress' ),
+					array(
+						'status' => 404,
+					)
+				);
```
renatonascalves, line 210: The same suggestion. I think this should be checked first.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -204,19 +204,20 @@ public function get_items( $request ) {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return true|WP_Error
+	 * @return boolean
```
renatonascalves, line 207: So, I actually think this change is incorrect. I could filter this hook and return an `WP_Error` object. The phpDoc is not only about what is used in the method but what it expects to be returned.

An intelephense tool, for example, will wrongly communicate this expects only a `boolean` when we know it doesn't.

Besides, `false` is the incorrect response for an error in a REST response. See the other examples where we are using a `WP_Error` object, not a `false` `boolean`.

renatonascalves: It's either `true` or `WP_Error`.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -268,18 +269,21 @@ public function get_item_permissions_check( $request ) {
...
+			$group  = $this->get_group_object( $request );
+
+			if ( empty( $group->id ) ) {
+				$retval = new WP_Error(
+					'bp_rest_group_invalid_id',
+					__( 'Invalid group ID.', 'buddypress' ),
+					array(
+						'status' => 404,
+					)
+				);
```
renatonascalves, line 283: The same suggestion. I think this should be checked first.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
```diff
@@ -1067,6 +1071,10 @@ protected function can_user_delete_or_update( $group ) {
 	 */
 	protected function can_see( $group ) {
 
+		if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
```
renatonascalves, line 1074: Looks like you are duplicating this check here and on line 273.

includes/bp-members/classes/class-bp-rest-attachments-member-avatar-endpoint.php
```diff
@@ -160,17 +160,27 @@ public function get_item( $request ) {
...
-		);
-		$this->user = bp_rest_get_user( $request->get_param( 'user_id' ) );
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'members' ) ) ) {
+			$retval     = new WP_Error(
+				'bp_rest_member_invalid_id',
+				__( 'Invalid member ID.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
```
renatonascalves, line 170: The same suggestion. I think this should be checked first.

includes/bp-members/classes/class-bp-rest-attachments-member-cover-endpoint.php
```diff
@@ -150,17 +150,27 @@ public function get_item( $request ) {
...
-		);
-		$this->user = bp_rest_get_user( $request->get_param( 'user_id' ) );
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'members' ) ) ) {
+			$retval     = new WP_Error(
+				'bp_rest_member_invalid_id',
+				__( 'Invalid member ID.', 'buddypress' ),
+				array(
+					'status' => 404,
+				)
+			);
```
renatonascalves, line 160: The same suggestion. I think this should be checked first.

includes/bp-members/classes/class-bp-rest-members-endpoint.php
```diff
@@ -198,19 +198,20 @@ public function get_items( $request ) {
 	 * @since 0.1.0
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
-	 * @return true
+	 * @return boolean
```
renatonascalves, line 201: See my previous comment, the `boolean` `true` here without the `WP_Error` might've been an oversight on my part.

renatonascalves (changes requested):
Left a few considerations!

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: This is a good question. I ordered the checks in this way because I thought if you didn't have the capability to access groups, then you shouldn't know whether group ID `x` resolved to a real group (to prevent fishing). I think this logic is similar to our hidden group logic. I am happy to be talked out of it, though, if that is the wrong way to approach a REST interaction. Thanks for reviewing this PR!

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
dcavins: I agree that this seems duplicative, but couldn't imagine leaving the `bp_view` check out of a function called `can_see()`. :) Happy to remove this if you prefer.

includes/bp-groups/classes/class-bp-rest-groups-endpoint.php
dcavins: Thanks for the explanation! I'll change this back everywhere I made that error.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: Yes, the `bp_view` capability is invoked in the `pre_query` filter, before we know if the request would really resolve. https://github.com/buddypress/buddypress/blob/master/src/bp-core/classes/class-bp-component.php#L1265
@imath Do you agree with this statement?
Thanks again!

renatonascalves: In that case, I'd suggest following BP core and leaving as is. Maybe returning the proper error code so that consumers of this API can handle it properly in their apps.

imath: @dcavins @renatonascalves It happens just after we have set the BP Queried object, firstly to avoid WordPress to query for posts for nothing as we are in a BuddyPress area. The `BP_Component::pre_query()` method filters `posts_pre_query`. This means if a group or a user wasn't found, we'd get a 404 before reaching the login form. I think @dcavins logic: no access, no need to check further would be better, but as we need to make sure we're in a BuddyPress context it's a bit different than group access (where we know we're already in BuddyPress).

If we would show the login form instead of the 404, as we redirect to the requested page after login, we'd redirect to a 404 which is a bit annoying.

On the other hand, this is something we'll need to discuss about again: because in terms of "security" it could be annoying if I wanted to build a _super_ private community as showing a 404 before the login form means "I confirm something exists if you see this form...".

dcavins: I agree that no matter which order we do the checks in, someone will be unhappy with that decision. :) But I think that we should choose security/privacy over convenience, generally.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: This is an interesting question. I agree that I've been thinking too narrowly that if you don't have the `bp_view` capability you need to log in, but that only applies to the basic private site logic. 

The function `rest_authorization_required_code(`) does try to manage this situation, though, by sending the right code, depending on the user context: 401 if the user is not logged in, 403 if the user is logged in. https://developer.wordpress.org/reference/functions/rest_authorization_required_code/

Did WP save us a decision here?

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
imath: I agree, 12.0.0-beta1 is tomorrow, this is something we can probably work on during the next 3 months. My first idea would be to include a new argument to `bp_do_404()` like $component, check cap inside it to eventually redirect to the component's directory.

renatonascalves: Sounds good. And I agree with the approach taken.

I do agree that some people might complain whatever approach is taken since I might be logged in already and if I deactivate access in one component, it'll redirect to the login page, even when I'm logged in. That'd be a weird user experience. I presume you'll provide ways to handle that in a non-REST context.

Here, the consumer of the API will have to handle these scenarios. :)

> I think that we should choose security/privacy over convenience, generally.

But it is odd that you see this as a security _thing_. If I visit `site.com/renato` and the page 404s, what's the security implication in this page not being found? Why that would be any different for a BuddyPress document?!

Even user IDs are exposed in the, WordPress, REST API. So as far as I know, there is no security or privacy implications here.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
renatonascalves: If I understood this feature correctly, I can disable access to specific components, make them private. So it is technically possible to be logged in, having access to the app, and try to access an endpoint that's disabled. So the response would communicate the wrong response/code.

imath:
rest_authorization_required_code() is what WordPress uses when a user is not allowed to view a post. If it's not the right code, could you tell us what is the code to use ?

imath
https://imathi.eu
________________________________
De : Renato Alves ***@***.***>
Envoyé : Monday, July 31, 2023 3:29:37 PM
À : buddypress/BP-REST ***@***.***>
Cc : imath ***@***.***>; Mention ***@***.***>
Objet : Re: [buddypress/BP-REST] Support the bp_view capability. (PR #487)


@renatonascalves commented on this pull request.

________________________________

In includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php<https://github.com/buddypress/BP-REST/pull/487#discussion_r1279298730>:

> @@ -169,17 +169,27 @@ public function get_item( $request ) {
         * @return true|WP_Error
         */
        public function get_item_permissions_check( $request ) {
-               $retval      = new WP_Error(
-                       'bp_rest_group_invalid_id',
-                       __( 'Invalid group ID.', 'buddypress' ),
-                       array(
-                               'status' => 404,
-                       )
-               );
-               $this->group = $this->groups_endpoint->get_group_object( $request );
+               if ( ! bp_current_user_can( 'bp_view', array( 'bp_component' => 'groups' ) ) ) {
+                       $retval = new WP_Error(
+                               'bp_rest_authorization_required',
+                               __( 'Sorry, you cannot view group details.', 'buddypress' ),
+                               array(
+                                       'status' => rest_authorization_required_code(),


Just to confirm: in a REST context, that function is returning the wrong error code based on the status of the app. The consumer of the API will handle that but we are communicating the wrong code for this feature.

We can use that as well if we want but I'm just highlighting it so that we all understand it is not the correct code for the feature.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: Hi @renatonascalves, Thanks for your feedback! I'm sorry, I'm not understanding you. Here are some scenarios and the response codes that are received. Can you check that these make sense?

When the community visibility is set to "anyone" and I make an unauthenticated GET request to http://bpdevel.local/wp-json/buddypress/v1/members/1, I get a 200 response with the member info in the body.

When visibility is set to "members" and I resend the request (unauthenticated), I get 401 Unauthorized back.

If the user is logged in, I get a 200.

We're not limiting by component yet, but the framework to do so is in place, so if I add some custom code to prevent access to the members component for this user, using a filter like this:

```
add_filter( 'bp_map_meta_caps', 'prevent_non_admin_user_from_viewing_members_component', 15, 4 );
function prevent_non_admin_user_from_viewing_members_component( $caps, $cap, $user_id, $args ) {
	if ( 'bp_view' === $cap ) {
		$component = isset( $args[0]['bp_component'] ) ? $args[0]['bp_component'] : '';
		// Only 'admin' can view the memebers component.
		if ( 'members' === $component && 1 !== $user_id ) {
			$caps = array( 'do_not_allow' );
		}
	}
	return $caps;
}
```

then my logged-in user who can't view this item gets a 403 Forbidden response.

includes/bp-xprofile/classes/class-bp-rest-xprofile-field-groups-endpoint.php
```diff
@@ -216,16 +216,17 @@ public function get_items( $request ) {
 	 * @return true|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
+		$retval = bp_current_user_can( 'bp_view', array( 'bp_component' => 'xprofile' ) );
```
imath, line 220: Shouldn't we return a `rest_authorization_required_code()` WP Error instead ?

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -189,16 +189,17 @@ public function get_items( $request ) {
 	 * @return true|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
+		$retval = bp_current_user_can( 'bp_view', array( 'bp_component' => 'xprofile' ) );
```
imath, line 193: Shouldn't we return a `rest_authorization_required_code()` WP Error instead ?

includes/bp-xprofile/classes/class-bp-rest-xprofile-fields-endpoint.php
```diff
@@ -267,16 +268,17 @@ public function get_item( $request ) {
 	 * @return true|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
+		$retval = bp_current_user_can( 'bp_view', array( 'bp_component' => 'xprofile' ) );
```
imath, line 272: Shouldn't we return a `rest_authorization_required_code()` WP Error instead ?

dcavins:
@imath You are correct about this:

> If you're logged in you have access to the endpoint, only not logged in users don't have access.
> The feature is: all that was publicly viewable by anyone becomes only viewable by logged in members.

but @renatonascalves makes an excellent point that visibility will be used in ways we do not expect. One example I can think of is for a subscription-based community site where, when your subscription expires, you can't view the community anymore (you no longer have the `bp_view` capability), whether you are logged in or not. So it's good to be thinking about this feature from different angles.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: Ah, are you talking about the `code` in this response and not the HTTP status code?

```
{
    "code": "bp_rest_authorization_required",
    "message": "Sorry, you are not allowed to perform this action.",
    "data": {
        "status": 403
    }
}
```

dcavins:
I think I finally understand that maybe Renato isn't talking about the HTTP status error code, but the code description, like `"code": "bp_rest_authorization_required",` and if that is the case, I can fix it! :)

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
dcavins: For instance, we could do something like this?
```
		$retval = new WP_Error(
			( is_user_logged_in() ) ? 'rest_forbidden' : 'bp_rest_authorization_required',
			__( 'Sorry, you are not allowed to perform this action.', 'buddypress' ),
			array(
				'status' => rest_authorization_required_code(),
			)
		);
```
Or should the error description/id/code be more specific, like `bp_forbidden_cannot_view_{$component}`?

It looks like if this is the correct way to handle mixed permissions, we have a number of places to update this, for instance https://github.com/buddypress/BP-REST/blob/master/includes/bp-members/classes/class-bp-rest-members-endpoint.php#L328

Thanks!

renatonascalves:
Maybe we should bring this conversation to Slack?

@dcavins The code I'm referring into is the HTTP code: 401 versus 403. This private feature is a flag feature and I was trying to call attention that those two error codes are not the true representation of this feature. If the site, or parts of it, is private, does it mean the user is logged out? That code, `bp_rest_authorization_required`, could be improved as well.

Some of my comments comes from some of the notes from the [ticket](https://buddypress.trac.wordpress.org/ticket/8734) and this screenshot: https://buddypress.trac.wordpress.org/attachment/ticket/8734/community-visibility-settings.png

I'm not sure if BP will start with a simple feature and expand it later. But based on what I read, it'd be technically possible to have access to the Activity component and be blocked from the Groups component.

From a REST context, in a app, the user is logged in. Requests are sent using tokens or cookies, for example, So we need a better way to communicate to the app that the component (the endpoint in this case) is blocked, not that the user is logged out, but that this is a feature flag which is blocking the request and the current user does not have access to it.

@imath In reply to your point [here](https://github.com/buddypress/BP-REST/pull/487#issuecomment-1658426066). ^ That's the reason that function does not communicate the true value. WordPress is using that function because it checks the user permissions, not if a feature flag is active.

That doesn't mean we can't use it for this purpose. I'm actually not even sure another HTTP error code exists for feature flags. Something to look into.

includes/bp-groups/classes/class-bp-rest-attachments-group-avatar-endpoint.php
renatonascalves: I replied it here: https://github.com/buddypress/BP-REST/pull/487#issuecomment-1658538394

imath:
Hi @renatonascalves & @dcavins 

Is there any requested change that wasn't implemented yet? Or are we good to merge this PR ?

I see the review requested change is still in red.

renatonascalves (approved):

renatonascalves:
sorry, all good to me.

imath:
Thanks a lot @renatonascalves 👍. I'm merging the changes. Many thanks to @dcavins for the great work.

PR #495, opened 2024-03-05
Use the versioned `buddypress/bp-coding-standards` from packagist

renatonascalves:
I'll be making some changes to the `dev-trunk` soon. So to avoid any issues here, I'm changing it to install from the stable version.

Related to https://github.com/buddypress/buddypress/pull/248


imath:
@renatonascalves Just did! Thanks for working on this improvement 👍

PR #501, opened 2024-04-16
Fix a DB Error when BP Blogs component is active on a regular WP

imath:
Steps to reproduce the DB error:

- Use a regular WP config (not Multisite)
- Activate Activity component if not already active 
- Activate the Blogs component if not already active
- Publish a new post (it should generate a `new_blog_post` activity
- Go to `buddypress/v1/activity?_embed=1` 

You should get the DB Error, I believe you can also get it trying to get the root blog using the corresponding endpoint.

I believe it’s only impacting the BP REST plugin as the Blogs REST Controller is only loaded when Multisite is active and a stable BuddyPress release is used. See https://github.com/buddypress/buddypress/blob/ef9fde9191b87a2f6a4027f2ecc75cfc1c616090/src/bp-blogs/classes/class-bp-blogs-component.php#L508

### DIFF
--- a/bp-rest.php
+++ b/bp-rest.php
@@ -86,7 +86,7 @@ function bp_rest() {
 		$controller->register_routes();
 	}
 
-	if ( bp_is_active( 'blogs' ) ) {
+	if ( is_multisite() && bp_is_active( 'blogs' ) ) {
 		require_once dirname( __FILE__ ) . '/includes/bp-blogs/classes/class-bp-rest-blogs-endpoint.php';
 		$controller = new BP_REST_Blogs_Endpoint();
 		$controller->register_routes();
--- a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
+++ b/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php
@@ -1149,7 +1149,7 @@ class BP_REST_Activity_Endpoint extends WP_REST_Controller {
 		}
 
 		// Embed Blog.
-		if ( bp_is_active( 'blogs' ) && buddypress()->blogs->id === $activity->component && ! empty( $activity->item_id ) ) {
+		if ( is_multisite() && bp_is_active( 'blogs' ) && buddypress()->blogs->id === $activity->component && ! empty( $activity->item_id ) ) {
 			$links['blog'] = array(
 				'embeddable' => true,
 				'href'       => rest_url(
--- a/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
+++ b/includes/bp-notifications/classes/class-bp-rest-notifications-endpoint.php
@@ -692,7 +692,7 @@ class BP_REST_Notifications_Endpoint extends WP_REST_Controller {
 		}
 
 		// Embed Blog.
-		if ( bp_is_active( 'blogs' ) && buddypress()->blogs->id === $notification->component_name && ! empty( $notification->item_id ) ) {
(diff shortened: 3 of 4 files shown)


renatonascalves:
🤔 I can't replicate this, btw. Tested on a fresh install with BP 12.4.0.

What's the DB error you see?

renatonascalves:
Actually, I forgot to add the `_embed` param. I see the right response, but also a `rest_no_route` error.

> "No route was found matching the URL and request method."

imath:
Interesting 🧐.

First, if you use the 12.4.0 BP stable version, I'm not sure the BP REST plugin is used even if activated, I believe it's using the built REST endpoints of BuddyPress.

The fact a regular endpoint ends up having no route is very weird, have you customized the `activity` component ID?

If I've found the issue with the activity endpoint, I think you can more easily get it trying to reach any Blogs component endpoint making sure the component is active and BuddyPress is activated on a NON Multisite config. A regular WP.

This endpoint's goal is to list blogs, it shouldn't be loaded in regular configs because there's only one blog. The BP REST plugin unlike BuddyPress is loading it... The failing unit tests are proving it.

renatonascalves:
Give me a minute to test it again. I was not using the BP REST plugin.

imath:
Here's the error I get:

```sh
[16-Apr-2024 17:15:44 UTC] WordPress database error Table 'develop_buddypress.wp_blogs' doesn't exist for query 
			SELECT b.blog_id, b.user_id as admin_user_id, u.user_email as admin_user_email, wb.domain, wb.path, bm.meta_value as last_activity, bm_name.meta_value as name
			FROM
			  wp_bp_user_blogs b
			  LEFT JOIN wp_bp_user_blogs_blogmeta bm ON (b.blog_id = bm.blog_id)
			  LEFT JOIN wp_bp_user_blogs_blogmeta bm_name ON (b.blog_id = bm_name.blog_id)
			  LEFT JOIN wp_bp_user_blogs_blogmeta bm_description ON (b.blog_id = bm_description.blog_id)
			  LEFT JOIN wp_blogs wb ON (b.blog_id = wb.blog_id)
			  LEFT JOIN wp_users u ON (b.user_id = u.ID)
			WHERE
			  wb.archived = '0' AND wb.spam = 0 AND wb.mature = 0 AND wb.deleted = 0 
			  AND bm.meta_key = 'last_activity' AND bm_name.meta_key = 'name' AND bm_description.meta_key = 'description'
			     AND b.blog_id IN (1) 
			GROUP BY b.blog_id ORDER BY bm.meta_value DESC  LIMIT 0, 20
		 made by require_once('wp-admin/admin-header.php'), do_action('admin_enqueue_scripts'), WP_Hook->do_action, WP_Hook->apply_filters, bp_admin_enqueue_scripts, do_action('bp_admin_enqueue_scripts'), WP_Hook->do_action, WP_Hook->apply_filters, bp_activity_admin_enqueue_assets, array_reduce, rest_preload_api_request, WP_REST_Server->response_to_data, WP_REST_Server->embed_links, WP_REST_Server->dispatch, WP_REST_Server->respond_to_request, BP_REST_Blogs_Endpoint->get_item, BP_REST_Blogs_Endpoint->get_blog_object, bp_blogs_get_blogs, BP_Blogs_Blog::get
```

renatonascalves (approved):
Confirmed! Makes sense. We are trying to get a blog without having any available.

Fix looks good!

PR #502, opened 2024-04-20
Registration through REST API with additional profile fields

### LINKED ISSUE #489: Allow additional profile fields to be included into the signup process
I agree with this ticket, we should mimic what we are making available from front end from the BP REST API.

See: https://buddypress.trac.wordpress.org/ticket/8976

renatonascalves:
Ticket: https://buddypress.trac.wordpress.org/ticket/8976

fixes #489

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-signup-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-signup-endpoint.php
@@ -13,8 +13,8 @@ defined( 'ABSPATH' ) || exit;
  *
  * Use /signup
  * Use /signup/{id}
- * Use /signup/resend/
- * Use /signup/activate/{id}
+ * Use /signup/resend
+ * Use /signup/activate/{activation_key}
  *
  * @since 6.0.0
  */
@@ -62,8 +62,9 @@ class BP_REST_Signup_Endpoint extends WP_REST_Controller {
 			array(
 				'args'   => array(
 					'id' => array(
-						'description' => __( 'Identifier for the signup. Can be a signup ID, an email address, or an activation key.', 'buddypress' ),
-						'type'        => 'string',
+						'description'       => __( 'Identifier for the signup. Can be a signup ID, an email address, or an activation key.', 'buddypress' ),
+						'type'              => 'string',
+						'sanitize_callback' => 'sanitize_text_field',
 					),
 				),
 				array(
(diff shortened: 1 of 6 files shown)


includes/bp-members/classes/class-bp-rest-signup-endpoint.php
```diff
@@ -402,10 +396,99 @@ public function create_item( $request ) {
 		// Hash and store the password.
 		$meta['password'] = wp_hash_password( $password );
 
-		$user_name = $request->get_param( 'user_name' );
```
imath, line 405: I'm not sure how deprecated fields are handled within the REST API, but I would probably keep this part to check if a `user_name` was sent within the request to eventually log a deprecated notice.

Then a bit later I'd check if the new way of getting signup fields generated a not empty $meta['field_1']. If $meta['field_1'] is empty, I'd probably try to set it using the `$user_name`.

imath (approved):
Hi @renatonascalves 

Thanks for your great work about it. I think we should go the way you improved the signup endpoint as it's consistent with what's happening on the web version.

I believe we simply need to make sure deprecated usage is traced for the developer so that he can update his code. That's why I added my inline comment as I'm unsure how the REST API is managing deprecated fields.

Once you committed, you can also write a developer note about the changes on bpdevel.wordpress.com to early inform about this coming change in BuddyPress 14.0

includes/bp-members/classes/class-bp-rest-signup-endpoint.php
renatonascalves: If the data is not sent, or it it sent empty, without `user_name`, the REST API request will fail. Meaning that there is no possibility for the data to fallback to `user_name`.

Any required field will fail a request, just like in the web version. So the user will either to fix their request to add it and/or to remove the `user_name` field.

renatonascalves: The only way out for consumers here is to migrate to `signup_field_data`.

renatonascalves: If you look at how the code is setup, having to account for this field as a fallback would be more trouble than it is worth, IMO. And we would lose the main goal here: have parity with the web version.

renatonascalves:
> Once you committed, you can also write a developer note about the changes on bpdevel.wordpress.com to early inform about this coming change in BuddyPress 14.0

I'll do this. 💪🏾

includes/bp-members/classes/class-bp-rest-signup-endpoint.php
renatonascalves: Done: https://bpdevel.wordpress.com/2024/05/07/signup-fields-via-the-rest-api/

PR #429, opened 2021-11-07
Paginating Thread messages/recipients

renatonascalves:
fixes #409 

Related: https://buddypress.trac.wordpress.org/ticket/8597

The BP REST API Documentation will need to be updated as well.

PR #503, opened 2024-04-21
Messages: add support for paginating thread messages

### LINKED ISSUE #409: Private messages pagination
I don't see how to paginate large conversations in a Thread. All messages are included in the thread, and i don't think it will scale, especially in a single thread conversation app.

renatonascalves:
Initially, at #429, we added support for message pagination while listing threads. In this pr, we add support when getting a single thread, and its messages.

fixes #409 
Ticket: https://buddypress.trac.wordpress.org/ticket/9146

### DIFF
--- a/includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
+++ b/includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
@@ -62,6 +62,15 @@ class BP_REST_Messages_Endpoint extends WP_REST_Controller {
 			$this->namespace,
 			$thread_endpoint,
 			array(
+				'args'   => array(
+					'id' => array(
+						'description'       => __( 'ID of the thread.', 'buddypress' ),
+						'type'              => 'integer',
+						'required'          => true,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
+					),
+				),
 				array(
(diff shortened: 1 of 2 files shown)


berzki:
Hi all,

I apologise for adding my comment here, but I am unable to create a Trac ticket for this enhancement. But I felt I should submit my feedback/request somehow. Paging was added to private messages in version 14 for the REST API here which is awesome. Are we able to get paging for private message threads in the web version too? As the whole thread is loaded as one page, threads with hundreds of messages have loading delays and put a lot of strain on resources. Another thing to note is that when you are in active conversation with others, the page will need to be refreshed to see the new messages. This is fine but adds to the overall load as the thread gets larger and the need to scroll all the way to the bottom to see the new messages (quite time consuming on a mobile phone).

If I ask nicely, do you think we could get this in version 14 if it's not too much trouble? It is certainly a much-needed feature. :)

Lastly, thanks to all for continuously providing new features for the community to use!

renatonascalves:
@berzki Thanks! This is something we discussed internally a while ago. I created a ticket so that we can handle it soon: https://buddypress.trac.wordpress.org/ticket/9156

berzki:
Thanks for this @renatonascalves! I hope this will be for both the Legacy (which is what I am using) and the Nouveau template? I've never used the Nouveau template, so I don't know much about how it works. I apologise if I misunderstood the ticket! Lol. :)

PR #506, opened 2024-05-21
Review the `bp_view` capability support in endpoints

### LINKED ISSUE #496: Review the `bp_view` capability support in endpoints
After adding support for the GraphQL API, https://github.com/renatonascalves/wp-graphql-buddypress/pull/124, I'm pretty sure we are missing support in a few endpoints.

See #487

renatonascalves:
- fixes #496 
- adds unit tests
- add support for missing components, i.e., activity

### DIFF
--- a/includes/bp-members/classes/class-bp-rest-signup-endpoint.php
+++ b/includes/bp-members/classes/class-bp-rest-signup-endpoint.php
@@ -203,18 +203,16 @@ class BP_REST_Signup_Endpoint extends WP_REST_Controller {
 			array( 'status' => rest_authorization_required_code() )
 		);
 
-		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'members' ) ) ) {
-			$capability = is_multisite() ? 'manage_network_users' : 'edit_users';
+		$capability = is_multisite() ? 'manage_network_users' : 'edit_users';
 
-			if ( ! is_user_logged_in() ) {
-				$retval = new WP_Error(
-					'bp_rest_authorization_required',
-					__( 'Sorry, you need to be logged in to perform this action.', 'buddypress' ),
-					array( 'status' => rest_authorization_required_code() )
-				);
-			} elseif ( bp_current_user_can( $capability ) ) {
-				$retval = true;
-			}
+		if ( ! is_user_logged_in() ) {
+			$retval = new WP_Error(
+				'bp_rest_authorization_required',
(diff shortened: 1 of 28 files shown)


renatonascalves:
Just a note that the GraphQL API, https://github.com/renatonascalves/wp-graphql-buddypress/pull/124, took a different approach. We don't return an error but rather an empty result. Mainly because in GraphQL, fields can be nested ([connections](https://www.apollographql.com/blog/explaining-graphql-connections)).

includes/bp-members/classes/class-bp-rest-signup-endpoint.php
```diff
@@ -203,20 +203,18 @@ public function get_items_permissions_check( $request ) {
 			array( 'status' => rest_authorization_required_code() )
 		);
 
-		$capability = 'edit_users';
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'members' ) ) ) {
```
imath, line 206: I think this part is not necessary as we need a higher cap to get signups, the one you are using at line 207.

includes/bp-members/classes/class-bp-rest-signup-endpoint.php
```diff
@@ -272,19 +270,35 @@ public function get_item( $request ) {
...
-		$retval = $this->get_items_permissions_check( $request );
+		$retval = new WP_Error(
+			'bp_rest_authorization_required',
+			__( 'Sorry, you are not authorized to perform this action.', 'buddypress' ),
+			array( 'status' => rest_authorization_required_code() )
+		);
 
-		if ( ! is_wp_error( $retval ) ) {
-			$signup = $this->get_signup_object( $request->get_param( 'id' ) );
+		if ( bp_current_user_can( 'bp_view', array( 'bp_component' => 'members' ) ) ) {
```
imath, line 279: I think this part is not necessary as we need a higher cap to get signups, the one you are using at line 280.

imath (commented):
Hi @renatonascalves another great job on this one 💪 . I would only remove the `'bp_view'` check cap you added into the signups endpoint as anyway a higher cap is needed (`'edit_users'` or `'manage_network_users'`)

includes/bp-members/classes/class-bp-rest-signup-endpoint.php
renatonascalves: Good call.

PR #507, opened 2024-05-21
Use `groups_$action_member()` function to manage memberships

imath:
Working on https://buddypress.trac.wordpress.org/ticket/9158 I realized using `BP_Groups_Member->$action()` methods was not a great idea. We really need to use `groups_$action_member()` functions instead.

In https://github.com/buddypress/buddypress/pull/293 I'm updating these functions so that it's possible to use them from the REST API (the `bp_is_item_admin()` checks were restricting their use to the Web).

This makes sure cache is cleared, group activities are removed when needed etc...
See: https://github.com/buddypress/buddypress/blob/edd1d2e33a2a4d3ace8216e06ba846a36f452b88/src/bp-groups/bp-groups-cache.php#L391|L394


imath:
Failing tests will stop fail once https://github.com/buddypress/buddypress/pull/293 is committed!

renatonascalves (approved):
Looks good!

PR #509, opened 2024-05-28
Admin cannot be demoted or removed from a group when he's the only one left

renatonascalves:
Ticket: https://buddypress.trac.wordpress.org/ticket/9163

### DIFF
--- a/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
+++ b/includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
@@ -58,12 +58,6 @@ class BP_REST_Group_Membership_Endpoint extends WP_REST_Controller {
 			$this->namespace,
 			'/' . $this->rest_base . '/(?P<group_id>[\d]+)/members',
 			array(
-				'args'   => array(
-					'group_id' => array(
-						'description' => __( 'A unique numeric ID for the Group.', 'buddypress' ),
-						'type'        => 'integer',
-					),
-				),
 				array(
 					'methods'             => WP_REST_Server::READABLE,
 					'callback'            => array( $this, 'get_items' ),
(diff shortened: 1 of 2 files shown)


imath (commented):
Hi @renatonascalves 

Thanks a lot for you work on this. It's fine with me, but I would do the same kind of check when a site admin tries to remove the last group admin (directly here or in another PR).

I believe for this second case it's not behaving the right way into the Web version, so I'll give it a check.

imath:
@emaralive good point. Maybe we should improve the error message in this case: "Group needs 1 admin at least. Please promote a user before trying to demote/remove this group admin".

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -396,28 +370,35 @@ public function update_item( $request ) {
 		$role     = $request->get_param( 'role' );
 		$group_id = $group->id;
 
+		$bp_loggedin_user_id = bp_loggedin_user_id();
+
+		// Check if the user is a member of the group.
```
renatonascalves, line 375: Returns an early error for clarity when a user is not part of the group. For `unban` actions, we need to allow since banned users are not part of the group anymore.

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -427,19 +408,14 @@ public function update_item( $request ) {
...
-					)
+					array( 'status' => 500 )
 				);
 			}
 		}
 
 		// Get updated group member.
 		$group_member = new BP_Groups_Member( $user->ID, $group_id );
 
-		// Setting context.
```
renatonascalves, line 440: The API consumer should be setting this.

includes/bp-groups/classes/class-bp-rest-group-membership-endpoint.php
```diff
@@ -473,23 +449,23 @@ public function update_item( $request ) {
...
+		$error = new WP_Error(
 			'bp_rest_authorization_required',
 			__( 'Sorry, you are not allowed to perform this action.', 'buddypress' ),
-			array(
-				'status' => rest_authorization_required_code(),
-			)
+			array( 'status' => rest_authorization_required_code() )
 		);
+
+		$admin_error = new WP_Error(
```
renatonascalves, line 458: Improving the error message.

renatonascalves:
@emaralive and @imath I updated the pr with your suggestions. If you could take a look at the unit tests, that'd be great.

If you see a scenario I haven't tested, let me know so that we can improve it.

emaralive:
@renatonascalves,

Superb improvement. :+1:  BTW, someone might quibble over the placement (where it resides) of "**at least**" in the error message but, that won't be me.

imath (approved):
Hi @renatonascalves sorry it took me so long to check it again. Thanks a lot for your improvements, let's have this in!

PR #510, opened 2024-06-06
BP9170: "raw" property missing from subject/excerpt/message nodes

renatonascalves:
Ticket: https://buddypress.trac.wordpress.org/ticket/9170


imath (approved):
Thanks a lot for this improvement @renatonascalves 💪👍

PR #511, opened 2024-07-11
Unable to send "0" in messages through REST API 

renatonascalves:
Related: https://github.com/buddypress/buddypress/pull/330
Ticket: https://buddypress.trac.wordpress.org/ticket/9175


includes/bp-messages/classes/class-bp-rest-messages-endpoint.php
```diff
@@ -345,20 +345,20 @@ public function get_item_permissions_check( $request ) {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function create_item( $request ) {
-		// Setting context.
-		$request->set_param( 'context', 'edit' );
```
renatonascalves, line 349: The API consumer should be providing this.

